PR #5, opened 2016-10-21
Add CLI benchmark device selector

AdamHarries:
At present, there is no offical way in the SYCL spec of selecting a specific OpenCL device for an existing, compiled application. This PR provides a SYCL device selector for matching devices of a requested vendor and type, as well as a command line flag, `--device` for the benchmark applications for triggering the selector.

### DIFF
--- a/benchmarks/benchmark.h
+++ b/benchmarks/benchmark.h
@@ -70,8 +70,6 @@ struct benchmark_arguments {
     /* Match parameters */
     std::regex output_regex("--output");
     std::regex device_regex("--device");
-    // device_vendor = std::string(".*");
-    // device_type   = std::string("*");
     /* Check if user has specified any options */
     bool match = true;
     for (int i = 1; i < argc; i++) {
@@ -84,7 +82,7 @@ struct benchmark_arguments {
       }
       // Check for the --output parameter
       if (std::regex_match(option, output_regex)) {
-        if (i + 1 >= argc) {
+        if ((i + 1) >= argc) {
           std::cerr << " Incorrect parameter " << std::endl;
           match = false;
           break;
@@ -108,7 +106,7 @@ struct benchmark_arguments {
 
       // Check for the --device parameter
       if (std::regex_match(option, device_regex)) {
-        if (i + 1 >= argc) {
+        if ((i + 1) >= argc) {
           std::cerr << " Incorrect parameter " << std::endl;
           match = false;
           break;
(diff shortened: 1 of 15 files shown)


benchmarks/benchmark.h
```diff
@@ -95,6 +106,36 @@ struct benchmark_arguments {
...
+          std::cerr << " Incorrect parameter " << std::endl;
+          match = false;
+          break;
+        }
+        std::string outputOption = argv[i + 1];
+        std::transform(outputOption.begin(), outputOption.end(),
+                       outputOption.begin(), ::tolower);
+        // split the string into tokens on ':'
+        std::stringstream ss;
+        ss.str(outputOption);
```
Ruyk: why don't use directly the constructor that takes an string?

benchmarks/sycl_exclusive_scan.cpp
```diff
@@ -41,25 +41,27 @@ using namespace sycl::helpers;
...
 
   auto exclusive_scan = [&]() {
-    cl::sycl::queue q;
+    cl::sycl::queue q(cds);
     sycl::sycl_execution_policy<class ExclusiveScanAlgorithm1> snp(q);
-    std::experimental::parallel::exclusive_scan(snp, begin(v1), end(v1), 
-      begin(v1), 0, [=](int x, int y){ return x + y; });
+    std::experimental::parallel::exclusive_scan(
+        snp, begin(v1), end(v1), begin(v1), 0,
+        [=](int x, int y) { return x + y; });
```
Ruyk: is this clang-format?

benchmarks/benchmark.h
```diff
@@ -61,6 +69,9 @@ struct benchmark_arguments {
         validProgramOptions(true) {
     /* Match parameters */
     std::regex output_regex("--output");
+    std::regex device_regex("--device");
+    // device_vendor = std::string(".*");
+    // device_type   = std::string("*");
```
AdamHarries: Ah, that was some changes I made while working out how to handle wildcards, and where the `--device` specifier isn't given. I'll remove them.

benchmarks/benchmark.h
```diff
@@ -95,6 +106,36 @@ struct benchmark_arguments {
         i++;
       }
 
+      // Check for the --device parameter
+      if (std::regex_match(option, device_regex)) {
+        if (i + 1 >= argc) {
```
AdamHarries: WIll do.

benchmarks/sycl_exclusive_scan.cpp
AdamHarries: I think it probably is - I ran clang-format over all the files that I'd changed before committing.

benchmarks/benchmark.h
AdamHarries: No reason!

PR #8, opened 2016-11-09
Replace all 3D ndrange with 1D range

### LINKED ISSUE #6: Error when compiling test file count_if.cpp
Hello,

When I tried to compile the tets file count_if.cpp, 'make' raises me an error at compile time which make me think that there is an implicit conversion from nd_item<1> to nd_item<3> in the file:
https://github.com/KhronosGroup/SyclParallelSTL/blob/master/include/sycl/algorithm/count_if.hpp#L88
which is likely to be a call to the constructor, located in this file :
https://github.com/KhronosGroup/SyclParallelSTL/blob/master/include/sycl/algorithm/algorithm_composite_patterns.hpp#L58

attached: 
The result from gcc when I try to compile count_if (attached below):
[make.log](https://github.com/KhronosGroup/SyclParallelSTL/files/579260/make.txt)
[count_if.cpp](https://github.com/KhronosGroup/SyclParallelSTL/files/579262/count_if.txt)
(this file is identical to https://github.com/KhronosGroup/SyclParallelSTL/blob/master/tests/count_if.cpp , except that I removed calls to GoogleTests)

Best regards

Joan Thibault

Ruyk:
Since all Parallel STL algorithms are currently working only with 1D
accessors, there is no need to use the 3D nd_range constructs.

This patch replaces all 3D nd_range and nd_items by the 1D counterparts. 

This should also fix #6

### DIFF
--- a/include/sycl/algorithm/algorithm_composite_patterns.hpp
+++ b/include/sycl/algorithm/algorithm_composite_patterns.hpp
@@ -51,11 +51,11 @@ class ReductionStrategy {
   int globalid_;
   int local_;
   int length_;
-  cl::sycl::nd_item<3> id_;
+  cl::sycl::nd_item<1> id_;
   local_rw_acc<T> scratch_;
 
  public:
-  ReductionStrategy(int loc, int len, cl::sycl::nd_item<3> i,
+  ReductionStrategy(int loc, int len, cl::sycl::nd_item<1> i,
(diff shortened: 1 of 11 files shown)


AdamHarries:
This all seems fine, builds and passes all tests on my setup

keryell:
Thank you. That looks good to me.
It should be more efficient too.

PR #9, opened 2016-11-09
update bodies printing to modern c++, fix memory leak

AdamHarries:
Trivial fix to the nbody benchmark to fix a (small) memory leak, and update the output code to be more modern c++


Ruyk:
@keryell good point with the const, but just found that didn't compile locally with ComputeCpp CE 0.1.1. I've reported the problem to ComputeCpp. In the meantime, I'll just accept this and we can add the const later :innocent:

PR #10, opened 2016-11-11
fix sycl_sample_01.cpp

JoanThibault:
replace call to b.set_final_data(v.data()) as buffer.set_final_data(T*) is undefined by:
construction of a shared_pointer to v (with no null_deleter)
creation of a buffer from this shared_pointer and a range

### DIFF
--- a/examples/sycl_sample_01.cpp
+++ b/examples/sycl_sample_01.cpp
@@ -59,14 +59,16 @@ class multiply_by_factor {
  * Note that for the moment the sycl variants of the algorithm
  *   are on the sycl namespace and not in std::experimental.
  */
+void null_deleter(void const *) {};
+
 int main() {
   std::vector<int> v = {3, 1, 5, 6};
   sycl::sycl_execution_policy<> sycl_policy;
 
   {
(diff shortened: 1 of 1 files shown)


keryell:
Thank you for the merge request.
Some minor points: there are some identation issues and I would prefer a lambda instead of a function for the `null_deleter`.
While you are at it you can also fix some typos in the comments that were already here.

But actually, since now you are no longer using `set_final_data`, you could just use
`cl::sycl::buffer<int> b(v.data(), cl::sycl::range<1>(v.size()));`
and the data will be copy back the data at destruction.
Could you update your branch towards this direction?
But we really need to improve buffer constructors in the spec to avoid this inconvenience.

Ruyk:
Since there are no ways of identifying a contiguous container in C++, maybe we can add a constructor to the SYCL buffer that takes explicitly an std::vector ?

Something like:
```
buffer<int> bufA(vec);
```
which will be roughly equivalent to:

```
buffer<int,1 > bufA(vec.data(), range<1>{vec.size()});
```

PR #12, opened 2016-11-28
upgrade: build.sh and CMakeLists.txt to be triSYCL compatible

JoanThibault:
build.sh is retro-compatible (i.e. ./build.sh /path/to/compute still works)
use ./build.sh --trisycl "path/to/trisycl" "path/to/compute" to use the trisycl mode
compute refers to boost/compute

CMakeLists.txt is also retro-compatible:
  set variables USE_COMUTECPP=OFF, TRISYCL_PACKAGE_ROOT_DIR and COMPUTE_PACKAGE_ROOT_DIR to use the triSYCL mode

Bonus:
    paths are made absolute
    the number of threads used by make is equal to the number of cores (returned by nproc)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -10,38 +10,38 @@ message(STATUS " Path to CMAKE source directory: ${CMAKE_SOURCE_DIR} ")
 set(CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/cmake/Modules/)
 
 if (USE_COMPUTECPP)
-	message(STATUS " Using ComputeCpp CMake")
+  message(STATUS " Using ComputeCpp CMake")
 
-	message(STATUS " Path to ComputeCpp implementation: ${COMPUTECPP_PACKAGE_ROOT_DIR} ")
+  message(STATUS " Path to ComputeCpp implementation: ${COMPUTECPP_PACKAGE_ROOT_DIR} ")
 
-	include(FindOpenCL)
-	include(FindComputeCpp)
+  include(FindOpenCL)
+  include(FindComputeCpp)
 
-	include_directories("${COMPUTECPP_INCLUDE_DIRECTORY}")
+  include_directories("${COMPUTECPP_INCLUDE_DIRECTORY}")
 
 else()
-	message(STATUS " Using triSYCL CMake")
-	set(CMAKE_CXX_STANDARD 14)
-	set(CXX_STANDARD_REQUIRED ON)
+  message(STATUS " Using triSYCL CMake")
+  set(CMAKE_CXX_STANDARD 14)
+  set(CXX_STANDARD_REQUIRED ON)
 
-	if(TRISYCL_INCLUDE_DIR)
-		message(STATUS "Path to triSYCL include directory: " ${TRISYCL_INCLUDE_DIR})
-		include_directories(${TRISYCL_INCLUDE_DIR})
-	endif()
-	if(BOOST_COMPUTE_INCLUDE_DIR)
-		message(STATUS "Path to Boost compute include directory: " ${BOOST_COMPUTE_INCLUDE_DIR})
-		include_directories(${BOOST_COMPUTE_INCLUDE_DIR})
-	endif()
+  if(TRISYCL_INCLUDE_DIR)
+    message(STATUS "Path to triSYCL include directory: " ${TRISYCL_INCLUDE_DIR})
+    include_directories(${TRISYCL_INCLUDE_DIR})
+  endif()
+  if(BOOST_COMPUTE_INCLUDE_DIR)
+    message(STATUS "Path to Boost compute include directory: " ${BOOST_COMPUTE_INCLUDE_DIR})
+    include_directories(${BOOST_COMPUTE_INCLUDE_DIR})
+  endif()
 
-	include(FindOpenCL)
+  include(FindOpenCL)
 
-	function(add_sycl_to_target targetName sourceFile binaryDir)
-	endfunction(add_sycl_to_target)
+  function(add_sycl_to_target targetName sourceFile binaryDir)
+  endfunction(add_sycl_to_target)
 
 endif()
 
 
-# PSTL specific 
+# PSTL specific
 include_directories("include")
 
 add_subdirectory (src)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,17 +4,42 @@ project (SyclSTL)
 enable_testing()
 
 option(PARALLEL_STL_BENCHMARKS "Build the internal benchmarks" OFF)
+option(USE_COMPUTECPP "Use ComputeCPP" ON)
 
-message(STATUS " Path to SYCL implementation: ${COMPUTECPP_PACKAGE_ROOT_DIR} ")
 message(STATUS " Path to CMAKE source directory: ${CMAKE_SOURCE_DIR} ")
-
-
 set(CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/cmake/Modules/)
 
-include(FindOpenCL)
-include(FindComputeCpp)
+if (USE_COMPUTECPP)
+	message(STATUS " Using ComputeCpp CMake")
 
-include_directories("${COMPUTECPP_INCLUDE_DIRECTORY}")
+	message(STATUS " Path to ComputeCpp implementation: ${COMPUTECPP_PACKAGE_ROOT_DIR} ")
+
+	include(FindOpenCL)
+	include(FindComputeCpp)
+
+	include_directories("${COMPUTECPP_INCLUDE_DIRECTORY}")
+
+else()
+	message(STATUS " Using triSYCL CMake")
+	set(CMAKE_CXX_STANDARD 14)
+	set(CXX_STANDARD_REQUIRED ON)
+
+	if(TRISYCL_INCLUDE_DIR)
+		message(STATUS "Path to triSYCL include directory: " ${TRISYCL_INCLUDE_DIR})
+		include_directories(${TRISYCL_INCLUDE_DIR})
+	endif()
+	if(BOOST_COMPUTE_INCLUDE_DIR)
+		message(STATUS "Path to Boost compute include directory: " ${BOOST_COMPUTE_INCLUDE_DIR})
+		include_directories(${BOOST_COMPUTE_INCLUDE_DIR})
+	endif()
+
+	include(FindOpenCL)
+
+	function(add_sycl_to_target targetName sourceFile binaryDir)
+	endfunction(add_sycl_to_target)
+
+endif()
+
 
 # PSTL specific 
 include_directories("include")
--- a/build.sh
+++ b/build.sh
@@ -1,13 +1,37 @@
 #!/bin/bash
-# the path to the ComputeCPP package root directory, e.g. /home/user/ComputeCpp-CE-0.1-Linux/
-PACKAGE_ROOT=$1
 
-#compute the number of cores
+# How to use build.sh to compile SyclParallelSTL with ComputeCpp ?
+# ./build.sh "path/to/ComputeCpp" (this path can be relative)
+#
+# for example:
+#	./build.sh /home/user/ComputeCpp
+#
+# How to use build.sh to compile SyclParallelSTL with triSYCL ?
+# ./build.sh --trisycl [-DTRISYCL_INCLUDE_DIR=path/to/triSYCL/include] [-DBOOST_COMPUTE_INCLUDE_DIR=path/to/boost/compute/include]
+#
+# for example (Ubuntu 16.04):
+#	./build.sh --trisycl -DTRISYCL_INCLUDE_DIR=~/triSYCL/include -DBOOST_COMPUTE_INCLUDE_DIR=~/compute/include
+#
+#
+
+
+# Useless to go on when an error occurs
+set -o errexit
+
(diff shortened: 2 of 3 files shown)


build.sh
```diff
@@ -1,6 +1,30 @@
...
+# ./build.sh "path/to/ComputeCpp"
+#
+# How to use build.sh to compile SyclParallelSTL with triSYCL ?
+# ./build.sh --trisycl "path/to/triSYCL" "path/to/compute"
+#
+# paths can be relative ones, this script will make them absoulte before to send them to CMake
+#
+# WARNING/TODO: setting the "path/to/compute" in triSYCL mode is not an option yet
+#
+
```
keryell: I would add:
```
# Useless to go on when an error occurs
set -o errexit

# Add error checking on number of args here and display typical use case
# ...
```

build.sh
```diff
@@ -1,6 +1,30 @@
...
+if [ $1 == "--trisycl" ]
+then
+	echo "build.sh enter mode: triSYCL"
+	CMAKE_ARGS="$CMAKE_ARGS -DUSE_COMPUTECPP=OFF -DTRISYCL_PACKAGE_ROOT_DIR=$(readlink -f $2)"
+	if [ $3 ]
```
keryell: Instead of dealing with this specific code, I would generalize and simplify all this test down to
```
CMAKE_ARGS+=" $@"
```
to pass all the other arguments down to CMake.

build.sh
```diff
@@ -1,6 +1,30 @@
...
+		CMAKE_ARGS="$CMAKE_ARGS -DCOMPUTE_PACKAGE_ROOT_DIR=$(readlink -f $3)"
+	fi
+else
+	echo "build.sh enter mode: ComputeCpp"
+	CMAKE_ARGS="$CMAKE_ARGS -DCOMPUTECPP_PACKAGE_ROOT_DIR=$(readlink -f $1)"
+fi
+NPROC=$(nproc)
 
 function install_gmock  {
   mkdir -p external && pushd external
```
keryell: The `git` afterwards should not fail if gmock is already there.
Typically because we run several times build.sh up to have it working... :-)

cmake/Modules/FindTriSYCL.cmake
```diff
@@ -0,0 +1,93 @@
...
+#	TRISYCL_PACKAGE_ROOT_DIR pointing to triSYCL
+#		available : https://github.com/keryell/triSYCL
+#	COMPUTE_PACKAGE_ROOT_DIR pointing to Compute
+#		available : https://github.com/boostorg/compute
+#
+
+# Require CMake version 3.2.2 or higher
+cmake_minimum_required(VERSION 3.2.2)
+
+# Check that a supported host compiler can be found
```
keryell: I am unsure all the following tests have any meaning for triSYCL

cmake/Modules/FindTriSYCL.cmake
```diff
@@ -0,0 +1,93 @@
...
+# Find Compute packagee
+if(NOT COMPUTE_PACKAGE_ROOT_DIR)
+	message(FATAL_ERROR "boost/compute package - Not found! (please set COMPUTE_PACKAGE_ROOT_DIR): ${COMPUTE_PACKAGE_ROOT_DIR}")
+else()
+	message(STATUS "boost/compute package - Found")
+endif()
+option(COMPUTE_PACKAGE_ROOT_DIR "Path to the boost/compute Package")
+
+# Obtain the triSYCL include directory
+set(TRISYCL_INCLUDE_DIRECTORY ${TRISYCL_PACKAGE_ROOT_DIR}/include/)
```
keryell: This makes me think that only `TRISYCL_INCLUDE_DIRECTORY` matters.

cmake/Modules/FindTriSYCL.cmake
```diff
@@ -0,0 +1,93 @@
...
+
+# Find triSYCL packagee
+if(NOT TRISYCL_PACKAGE_ROOT_DIR)
+	message(FATAL_ERROR "triSYCL package - Not found! (please set TRISYCL_PACKAGE_ROOT_DIR): ${TRISYCL_PACKAGE_ROOT_DIR}")
+else()
+	message(STATUS "triSYCL package - Found")
+endif()
+option(TRISYCL_PACKAGE_ROOT_DIR "Path to the triSYCL Package")
+
+# Find Compute packagee
```
keryell: It is not clear to me how it can works with pre-installed normal Boost.Compute.
Just rely on the normal CMake way to add directories to search for some include files?

build.sh
JoanThibault: currently it's not possible as git clone raises an error if googletest folder already  exists

build.sh
JoanThibault: the call to readlink allow to turn a relative adress into an absolute one (necessary as every piece of codes are not in the same folder)

JoanThibault: However, we could imagine a kind of '--cmake' tag which forecast every nexts options to cmake

build.sh
JoanThibault: I did an other PR on this subject as it was not triSYCL related.

build.sh
```diff
@@ -14,29 +17,50 @@
...
 
-function install_gmock  {
-  mkdir -p external && pushd external
-  git clone git@github.com:google/googletest.git
-  pushd googletest/googlemock/make && make
-  popd 
-  popd
-}
+function install_gmock  {(
+  #REPO="git@github.com:google/googletest.git"
```
Ruyk: Please remove the commented line

CMakeLists.txt
```diff
@@ -4,17 +4,40 @@ project (SyclSTL)
...
+	include(FindComputeCpp)
+
+	include_directories("${COMPUTECPP_INCLUDE_DIRECTORY}")
+
+else()
+	message(STATUS " Using triSYCL CMake")
 
-include(FindOpenCL)
-include(FindComputeCpp)
+	add_compile_options("-std=c++14")
```
Ruyk: this will break the windows build. 
The correct way of setting the compiler version is using the CXX standard property of the targets:

https://cmake.org/cmake/help/v3.1/prop_tgt/CXX_STANDARD.html

build.sh
```diff
@@ -1,23 +1,71 @@
...
+	fi
+else
+	echo "build.sh enter mode: ComputeCpp"
+	CMAKE_ARGS="$CMAKE_ARGS -DCOMPUTECPP_PACKAGE_ROOT_DIR=$(readlink -f $1)"
+	shift
+fi
+NPROC=$(nproc)
+
+function install_gmock  {(
+  #REPO="git@github.com:google/googletest.git"
```
Ruyk: Please remove the commented line

build.sh
```diff
@@ -1,23 +1,71 @@
...
+  make -j$NPROC
+)}
 
 function configure  {
-  mkdir -p build && pushd build 
-  cmake .. -DCOMPUTECPP_PACKAGE_ROOT_DIR=$PACKAGE_ROOT  -DPARALLEL_STL_BENCHMARKS=ON
+  mkdir -p build && pushd build
+  if [ $@ ]
+  then
+	  echo "cmake additionnal args: $@"
```
Ruyk: additionnal -> additional

cmake/Modules/FindTriSYCL.cmake
```diff
@@ -0,0 +1,93 @@
...
+
+# Check that a supported host compiler can be found
+if(CMAKE_COMPILER_IS_GNUCXX)
+    # Require at least gcc 4.8
+    if (CMAKE_CXX_COMPILER_VERSION VERSION_LESS 4.8)
+      message(FATAL_ERROR
+        "host compiler - Not found! (gcc version must be at least 4.8)")
+    # Require the GCC dual ABI to be disabled for 5.1 or higher
+    elseif (CMAKE_CXX_COMPILER_VERSION VERSION_GREATER 5.1)
+      set(COMPUTECPP_DISABLE_GCC_DUAL_ABI "True")
```
Ruyk: This is probably not needed on triSYCL since you are not distributing binaries built with the older version of the GCC ABI

CMakeLists.txt
```diff
@@ -4,17 +4,40 @@ project (SyclSTL)
 enable_testing()
 
 option(PARALLEL_STL_BENCHMARKS "Build the internal benchmarks" OFF)
+option(USE_COMPUTECPP "Use ComputeCPP" ON)
```
keryell: I am not sure we need this option.
What is the meaning of answering "OFF"?

CMakeLists.txt
```diff
@@ -4,17 +4,40 @@ project (SyclSTL)
...
+	message(STATUS " Using triSYCL CMake")
 
-include(FindOpenCL)
-include(FindComputeCpp)
+	add_compile_options("-std=c++14") 
+
+	set(TRISYCL_INCLUDE_DIR "/usr/include" CACHE STRING "Path to triSYCL/include")
+	set(BOOST_COMPUTE_INCLUDE_DIR  "/usr/include" CACHE STRING "Path to Boost-compute/include")
+
+	message(STATUS "Path to triSYCL/include = " ${TRISYCL_INCLUDE_DIR})
```
keryell: If we go this way, I guess it would be `"Path to triSYCL include directory="` sine /include has no meaning on some OS.

CMakeLists.txt
```diff
@@ -4,17 +4,40 @@ project (SyclSTL)
...
+	include_directories("${COMPUTECPP_INCLUDE_DIRECTORY}")
+
+else()
+	message(STATUS " Using triSYCL CMake")
 
-include(FindOpenCL)
-include(FindComputeCpp)
+	add_compile_options("-std=c++14") 
+
+	set(TRISYCL_INCLUDE_DIR "/usr/include" CACHE STRING "Path to triSYCL/include")
```
keryell: Adding `/usr/include` does not really make sense.
It is OS dependent.
Anyway what is required here is just nothing or something specified as a parameter when triSYCL path.

build.sh
```diff
@@ -1,23 +1,71 @@
...
+  fi
+  cmake .. $CMAKE_ARGS -DPARALLEL_STL_BENCHMARKS=ON $@
```
keryell: How is it possible to add more arguments to cmake without setting the triSYCL and Boost.Compute path?
The more I think to it and the less I am kind of adding all that stuff to specify where triSYCL and Boost.Compute are...

cmake/Modules/FindTriSYCL.cmake
```diff
@@ -0,0 +1,93 @@
...
+#
+#   TODO : add Copyright
+
+#########################
+#  FindTriSYCL.cmake  
+#########################
+#
+#  Tools for finding and building with triSYCL.
+#
+#  User must define:
```
keryell: Is it a must?

cmake/Modules/FindTriSYCL.cmake
```diff
@@ -0,0 +1,93 @@
...
+#	TRISYCL_INCLUDE_DIR pointing to triSYCL
+#		available : https://github.com/keryell/triSYCL
+#	BOOST_COMPUTE_INCLUDE_DIR pointing to Compute
+#		available : https://github.com/boostorg/compute
+#
```
keryell: OK, but then where are this variables used?

cmake/Modules/FindTriSYCL.cmake
```diff
@@ -0,0 +1,93 @@
...
+find_package(OpenCL REQUIRED)
+
+# Find triSYCL packagee
+if(NOT TRISYCL_INCLUDE_DIR)
+	message(FATAL_ERROR "triSYCL package - Not found! (please set TRISYCL_INCLUDE_DIR): ${TRISYCL_INCLUDE_DIR}")
+else()
+	message(STATUS "triSYCL package - Found")
+endif()
+option(TRISYCL_INCLUDE_DIR "Path to the triSYCL Package")
+
```
keryell: Why having an error if `TRISYCL_INCLUDE_DIR` is undefined?

cmake/Modules/FindTriSYCL.cmake
```diff
@@ -0,0 +1,93 @@
...
+# Find Compute packagee
+if(NOT BOOST_COMPUTE_INCLUDE_DIR)
+	message(FATAL_ERROR "boost/compute package - Not found! (please set BOOST_COMPUTE_INCLUDE_DIR): ${BOOST_COMPUTE_INCLUDE_DIR}")
+else()
+	message(STATUS "boost/compute package - Found")
+endif()
+option(BOOST_COMPUTE_INCLUDE_DIR "Path to the boost/compute Package")
+
```
keryell: Why having an error if `BOOST_COMPUTE_INCLUDE_DIR` is undefined?

CMakeLists.txt
JoanThibault: if you set this option to OFF you will use triSYCL instead, I agree that the variables name might not be the most meaningful.

Ruyk:
Seems the branch has conflicts that must be resolved before merging, can you do that @JoanThibault  ?

keryell:
That looks quite nicer now.
So now the ParallelSTL can use triSYCL!
This is ready to land on my side.
Thank you.

Ruyk:
Can you update the README with instructions on how to build with triSYCL ? That will make it simpler for people to build the project. With the README update this will be ready to go, good work!

Ruyk:
Quite fast, thanks!

PR #14, opened 2016-11-29
prevent failure when installing googletests

JoanThibault:
I replaced calls to pushd/popd by parenthesis
made the call to multi-threads make
calls git pull to keeps the googletest folder synced with the remote repository

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -2,6 +2,9 @@
 # the path to the ComputeCPP package root directory, e.g. /home/user/ComputeCpp-CE-0.1-Linux/
 PACKAGE_ROOT=$1
 
+#compute the number of cores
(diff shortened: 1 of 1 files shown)


build.sh
```diff
@@ -2,13 +2,32 @@
...
+      cd googletest
+      git pull
+    else
+      git clone $REPO
+      cd googletest
+    fi
+  else
+    mkdir external
+    cd external
+    git clone $REPO
```
keryell: I have the feeling the logic could be simplified a little bit here to avoid having the git clone command used twice. What about using BDD to minimize the code size here? :-)
Feel free to add comments.

keryell:
This looks good to me.
Thank you.

PR #15, opened 2016-11-29
simplify helpers/sycl_buffers.hpp

JoanThibault:
no write-back is the default behaviour of buffer's constructor called with two iterators

### DIFF
--- a/include/sycl/helpers/sycl_buffers.hpp
+++ b/include/sycl/helpers/sycl_buffers.hpp
@@ -104,12 +104,7 @@ template <typename Iterator,
 cl::sycl::buffer<typename std::iterator_traits<Iterator>::value_type, 1>
 make_buffer_impl(Iterator b, Iterator e, std::input_iterator_tag) {
   typedef typename std::iterator_traits<Iterator>::value_type type_;
-  size_t bufferSize = std::distance(b, e);
-  std::unique_ptr<type_> up{new type_[bufferSize]};
-  std::copy(b, e, up.get());
-  cl::sycl::buffer<type_, 1> buf(std::move(up), cl::sycl::range<1>(bufferSize));
-  buf.set_final_data(nullptr);
-  return buf;
+  return cl::sycl::buffer<type_, 1>(b, e);
 }
 
 /**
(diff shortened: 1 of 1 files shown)


include/sycl/helpers/sycl_buffers.hpp
```diff
@@ -104,12 +104,7 @@ template <typename Iterator,
 cl::sycl::buffer<typename std::iterator_traits<Iterator>::value_type, 1>
 make_buffer_impl(Iterator b, Iterator e, std::input_iterator_tag) {
   typedef typename std::iterator_traits<Iterator>::value_type type_;
-  size_t bufferSize = std::distance(b, e);
-  std::unique_ptr<type_> up{new type_[bufferSize]};
-  std::copy(b, e, up.get());
-  cl::sycl::buffer<type_, 1> buf(std::move(up), cl::sycl::range<1>(bufferSize));
-  buf.set_final_data(nullptr);
-  return buf;
+  return cl::sycl::buffer<type_, 1>(b, e);
```
keryell: Nice Ruyman. You found a bug in the SYCL specification. :-)
The SYCL 2.2 table 4.2 is incompatible with Section 4.2.2 item 7.
Let's discuss this tomorrow during the weekly SYCL committee meeting.
So no need to change the PSTL. Just fix the spec and update triSYCL.

Thanks.

JoanThibault: note: set_final_data(nullptr) has no defined behaviour either

keryell: @JoanThibault we solved the discrepancy in the specification in SYCL committee. The ComputeCpp choice is the more C++ friendly one. So you can improve the constructor in triSYCL.

Ruyk:
@keryell @JoanThibault  what do we do with this pull request? do we make the change with the if/defs ?

include/sycl/helpers/sycl_buffers.hpp
```diff
@@ -108,7 +108,9 @@ make_buffer_impl(Iterator b, Iterator e, std::input_iterator_tag) {
   std::unique_ptr<type_> up{new type_[bufferSize]};
   std::copy(b, e, up.get());
   cl::sycl::buffer<type_, 1> buf(std::move(up), cl::sycl::range<1>(bufferSize));
+#ifdef __COMPUTECPP__
```
keryell: OK, so ComputeCpp follows what I understand from the spec.
Great.
Just remove this ```buf.set_final_data(nullptr);```

keryell:
This sounds good!
The best code we write is always the code we do not need to write. :-)
Thanks.

Ruyk:
Thanks for the patch!

PR #18, opened 2016-12-07
reduce and hierarchical parallelism

JoanThibault:
alternative implementation of reduce using hierarchical parallelism.

### DIFF
--- a/include/sycl/algorithm/reduce.hpp
+++ b/include/sycl/algorithm/reduce.hpp
@@ -48,7 +48,7 @@ namespace impl {
  * Note that there is a potential race condition while using the same buffer for
  * input-output
  */
-#ifdef __COMPUTECPP__
+#if CL_SYCL_LANGUAGE_VERSION < 220
 template <class ExecutionPolicy, class Iterator, class T, class BinaryOperation>
 typename std::iterator_traits<Iterator>::value_type reduce(
     ExecutionPolicy &sep, Iterator b, Iterator e, T init, BinaryOperation bop) {
--- a/include/sycl/algorithm/reduce.hpp
+++ b/include/sycl/algorithm/reduce.hpp
@@ -45,9 +45,10 @@ namespace impl {
 /* reduce.
  * Implementation of the command group that submits a reduce kernel.
  * The kernel is implemented as a lambda.
- * Note that there is a potential race condition while using the same buffer fot
+ * Note that there is a potential race condition while using the same buffer for
  * input-output
  */
+#ifdef __COMPUTECPP__
 template <class ExecutionPolicy, class Iterator, class T, class BinaryOperation>
 typename std::iterator_traits<Iterator>::value_type reduce(
     ExecutionPolicy &sep, Iterator b, Iterator e, T init, BinaryOperation bop) {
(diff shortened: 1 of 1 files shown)


include/sycl/algorithm/reduce.hpp
```diff
@@ -94,6 +95,76 @@ typename std::iterator_traits<Iterator>::value_type reduce(
...
+
+			auto buff_acc = buff.template get_access<cl::sycl::access::mode::read_write>(cgh);
+
+			cl::sycl::accessor<type_, 1, cl::sycl::access::mode::read_write,
+			                             cl::sycl::access::target::local> scratch(cl::sycl::range<1>(nb_work_item), cgh);
+			cgh.parallel_for_work_group<typename ExecutionPolicy::kernelName>(rng, [=](cl::sycl::group<1> grp) {
+				
+				// Step 1:
+				//     copy a sub-vector of 'buff' into 'scratch'
+				grp.parallel_for_work_item([&](cl::sycl::nd_item<1> id) {
```
Ruyk: In the published SYCL 1.2 specification, parallel_for_work_item is a free function, not a method of the group. That is different on the provisional SYCL 2.2, but it is still provisional so we should keep Parallel STL according to the published official 1.2 specification.

include/sycl/algorithm/reduce.hpp
```diff
@@ -94,6 +95,76 @@ typename std::iterator_traits<Iterator>::value_type reduce(
                                      cl::sycl::access::target::host_buffer>();
   return hI[0] + init;
 }
+#else
+template <class ExecutionPolicy, class Iterator, class T, class BinaryOperation>
+typename std::iterator_traits<Iterator>::value_type reduce(
+    ExecutionPolicy &sep, Iterator b, Iterator e, T init, BinaryOperation bop) {
+
+	typedef typename std::iterator_traits<Iterator>::value_type type_;
```
keryell: Please do not use typedef in new code.

include/sycl/algorithm/reduce.hpp
```diff
@@ -94,6 +95,76 @@ typename std::iterator_traits<Iterator>::value_type reduce(
...
+			auto buff_acc = buff.template get_access<cl::sycl::access::mode::read_write>(cgh);
+
+			cl::sycl::accessor<type_, 1, cl::sycl::access::mode::read_write,
+			                             cl::sycl::access::target::local> scratch(cl::sycl::range<1>(nb_work_item), cgh);
+			cgh.parallel_for_work_group<typename ExecutionPolicy::kernelName>(rng, [=](cl::sycl::group<1> grp) {
+				
+				// Step 1:
+				//     copy a sub-vector of 'buff' into 'scratch'
+				grp.parallel_for_work_item([&](cl::sycl::nd_item<1> id) {
+					int globalid = id.get_global(0);
```
Ruyk: There is no get_global/get_local for item on SYCL 1.2, I am not sure it is possible to implement the reduction using Hierarchical with the current SYCL 1.2 without it?

include/sycl/algorithm/reduce.hpp
```diff
@@ -94,6 +95,76 @@ typename std::iterator_traits<Iterator>::value_type reduce(
...
   return hI[0] + init;
 }
+#else
+template <class ExecutionPolicy, class Iterator, class T, class BinaryOperation>
+typename std::iterator_traits<Iterator>::value_type reduce(
+    ExecutionPolicy &sep, Iterator b, Iterator e, T init, BinaryOperation bop) {
+
+	typedef typename std::iterator_traits<Iterator>::value_type type_;
+
+	cl::sycl::queue q(sep.get_queue());
```
keryell: Please write modern C++ code in new code.
```
cl::sycl::queue q { sep.get_queue() };
```

include/sycl/algorithm/reduce.hpp
```diff
@@ -94,6 +95,76 @@ typename std::iterator_traits<Iterator>::value_type reduce(
...
+
+	auto buff = sycl::helpers::make_const_buffer(b, e);
+	
+	size_t length = vectorSize;
+	while (length > 1) {
+		q.submit([&] (cl::sycl::handler &cgh) {
+			cl::sycl::nd_range<1> rng { cl::sycl::range<1>{std::max(length, nb_work_item)},
+										cl::sycl::range<1>{nb_work_item}};
+
+			auto buff_acc = buff.template get_access<cl::sycl::access::mode::read_write>(cgh);
```
keryell: If I understand correctly, you are using a read-write buffer to just get one element per work-group back at the end. I have the feeling it should be more efficient to have 1 read buff and 1 smaller write buffer to get the result.

include/sycl/algorithm/reduce.hpp
```diff
@@ -94,6 +95,76 @@ typename std::iterator_traits<Iterator>::value_type reduce(
...
+						int localid = id.get_local(0);
+						if (localid < offset) {
+							scratch[localid] = bop(scratch[localid], scratch[localid + offset]);
+						}
+					});
+				}
+
+				// Step 3:
+				//     copy back the reduction of 'scratch' into 'buff'
+				buff_acc[grp.get(0)] = scratch[0];
```
keryell: After discussion together, it looks like there was already this race condition in the original algorithm you translated... So I guess it worths to change the algorithm here, while moving to hierarchical parallelism.

include/sycl/algorithm/reduce.hpp
keryell: Ah, interesting! We need to verify it is on the list of the SYCL 1.2.1 update...

include/sycl/algorithm/reduce.hpp
```diff
@@ -45,9 +45,10 @@ namespace impl {
 /* reduce.
  * Implementation of the command group that submits a reduce kernel.
  * The kernel is implemented as a lambda.
- * Note that there is a potential race condition while using the same buffer fot
+ * Note that there is a potential race condition while using the same buffer for
  * input-output
  */
+#ifdef __COMPUTECPP__
```
keryell: OK, so to follow Ruyman proposal, use
```
#if CL_SYCL_LANGUAGE_VERSION < 220
```
instead

include/sycl/algorithm/reduce.hpp
Naghasan: To complement, the parallel_for_work_item takes an item not a nd_item (in both 1.2 and 2.2 specs)

include/sycl/algorithm/reduce.hpp
Naghasan: In the current spec, the user need to compute the global id him self. The group object and the local provide the required information to do so.

keryell:
I tried this version: ready to go on my side.

include/sycl/algorithm/reduce.hpp
keryell: Good point!
I have to update triSYCL on this.
I definitely do not want barriers in the parallel_for_work_item.

JoanThibault:
I did a new pull request, which integrates your remarks especially on local memory restrictions.

PR #21, opened 2017-01-31
Rewrite algorithms with hierarchical parallelism

JoanThibault:
I factorized algorithms into three main ones:
mapreduce
  => used to implement transform_reduce, count_if, fill, reduce
  NB: the binary operator is assumed to be associative and commutative
map2reduce (same as mapreduce, but takes two buffers as input instead of one)
  => used to implement inner_product
mapscan
  => used to implement inclusive_scan and exclusive_scan
  NB: the binary operator is assumed to be associative

They have been tested using the current implementation of triSYCL.


benchmarks/sycl_inclusive_scan.cpp
```diff
@@ -38,24 +38,271 @@
 
 using namespace sycl::helpers;
 
+size_t up_rounded_division(size_t x, size_t y){
```
keryell: Please fix all the formatting issues.

benchmarks/sycl_accumulate.cpp
```diff
@@ -0,0 +1,195 @@
...
+        }
+        output[group_id] = acc;
+      });
+    });
+    q.wait_and_throw();
+    auto read_output  = output_buff.template get_access
+      <cl::sycl::access::mode::read, cl::sycl::access::target::host_buffer>();
+    
+    int acc = 0;
+    for(size_t pos0 = 0; pos0 < nb_work_group; pos0++)
```
keryell: Why this is not a STL algorithm?

benchmarks/sycl_accumulate.cpp
```diff
@@ -0,0 +1,195 @@
...
+            //we peal the first iteration
+            T acc = input[local_pos];
+            for(size_t read = local_pos + nb_work_item; read < group_end; read += nb_work_item) {
+              acc = bop(acc, input[read]);
+            }
+            sum[local_id] = acc;
+          }
+        });
+        T acc = sum[0];
+        for(size_t local_id = 1; local_id < min(nb_work_item, group_end - group_begin); local_id++) {
```
keryell: Why this is not a STL algorithm? After all this is about the STL. :-)

benchmarks/sycl_accumulate.cpp
```diff
@@ -0,0 +1,195 @@
...
+        size_t group_begin = group_id * size_per_work_group;
+        size_t group_end   = min((group_id+1) * size_per_work_group, size);
+        assert(group_begin < group_end); //as we properly selected the number of work_group
+        grp.parallel_for_work_item([&](cl::sycl::nd_item<1> id) {
+          size_t local_id = id.get_local(0);
+          size_t local_pos = group_begin + local_id;
+          if (local_pos < group_end) {
+            //we peal the first iteration
+            T acc = input[local_pos];
+            for(size_t read = local_pos + nb_work_item; read < group_end; read += nb_work_item) {
```
keryell: Why this is not a STL algorithm? After all this is about the STL. :-)

benchmarks/sycl_accumulate.cpp
```diff
@@ -0,0 +1,195 @@
...
+          }
+        });
+        T acc = sum[0];
+        for(size_t local_id = 1; local_id < min(nb_work_item, group_end - group_begin); local_id++) {
+          acc = bop(acc, sum[local_id]);
+        }
+        output[group_id] = acc;
+      });
+    });
+    q.wait_and_throw();
```
keryell: Do we need to wait?

benchmarks/sycl_inclusive_scan.cpp
```diff
@@ -38,24 +38,271 @@
...
+  //return x / y + ((x % y == 0) ? 0 : 1);
+  return (x+(y-1)) / y;
+}
+
+typedef struct mapscan_descriptor {
+  size_t size, size_per_work_group, size_per_work_item, nb_work_group, nb_work_item;
+  mapscan_descriptor(size_t size_, size_t size_per_work_group_ , size_t size_per_work_item_, size_t nb_work_group_, size_t nb_work_item_):
+    size(size_),
```
keryell: Use modern { } syntax everywhere.

benchmarks/sycl_inclusive_scan.cpp
```diff
@@ -38,24 +38,271 @@
...
+  std::cout << "max_work_item=\t" << max_work_item << std::endl;
+  size_t nb_work_item = min(max_work_item, size_per_work_group);
+  std::cout << "nb_work_item=\t" << nb_work_item << std::endl;
+  size_t size_per_work_item = up_rounded_division(size_per_work_group, nb_work_item);
+  std::cout << "size_per_work_item=\t" << size_per_work_item << std::endl;
+  return mapscan_descriptor(size, size_per_work_group, size_per_work_item, nb_work_group, nb_work_item);
+}
+
+
+template <class ExecutionPolicy, class A, class B, class Reduce, class Map>
```
keryell: I prefer typename to class in templates.

build.sh
```diff
@@ -60,12 +61,14 @@ function mak  {
 function tst {
   pushd build/tests
   ctest -j$NPROC
-  popd
+  popd
```
keryell: Trailing space. How is configured your editor?

build.sh
```diff
@@ -1,10 +1,11 @@
 #!/bin/bash
 
 # How to use build.sh to compile SyclParallelSTL with ComputeCpp ?
-# ./build.sh "path/to/ComputeCpp" (this path can be relative)
+#   ./build.sh "path/to/ComputeCpp" (this path can be relative)
 #
 # for example:
-#  ./build.sh /home/user/ComputeCpp
+#	  ./build.sh /home/user/ComputeCpp
```
keryell: I guess the 2 ./build.sh lines should be indented the same way...

include/sycl/algorithm/fill.hpp
```diff
@@ -50,7 +50,9 @@ void fill(ExecutionPolicy &sep, ForwardIt b, ForwardIt e, const T &value) {
     auto device = q.get_device();
     size_t localRange =
         device.get_info<cl::sycl::info::device::max_work_group_size>();
-    auto bufI = sycl::helpers::make_buffer(b, e);
+    //typedef typename std::iterator_traits<ForwardIt>::value_type type_ ;
+    //cl::sycl::buffer<type_, 1> bufI { b, e };
+    auto bufI = helpers::make_buffer( b, e );
```
keryell: Then the spacing of this line shuld be the same as the original one

tests/inclusive_scan.cpp
```diff
@@ -48,10 +48,10 @@ void inclusive_scan_gold(std::vector<T>& v, T init = 0,
 
 // test the gold computation against a known value
 TEST_F(InclusiveScanAlgorithm, TestSTDInclusiveScan) {
-  std::vector<int> v = {5, 1, 6, 2, 6, 2, 5, 7};
+  std::vector<int> v    = {5, 1,  6,  2,  6,  2,  5,  7};
   std::vector<int> gold = {5, 6, 12, 14, 20, 22, 27, 34};
 
-  inclusive_scan_gold(v, 0, [](int a, int b) { return a + b; });
+  inclusive_scan_gold(v, 0, std::plus<int>());
```
keryell: The use of these operators make the code shorter. Nice!
Probably in modern C++ std::plus<int>{}

tests/inclusive_scan.cpp
```diff
@@ -65,7 +65,8 @@ TEST_F(InclusiveScanAlgorithm, TestSyclInclusiveScanPowerOfTwo1) {
   std::vector<int> v = {5, 1, 6, 2, 6, 2, 5, 7};
   std::vector<int> gold(v);
 
-  inclusive_scan_gold(gold, 0, [](int a, int b) { return a + b; });
+  inclusive_scan_gold(gold, 0, std::plus<int>());
+      //std::plus<int>());
```
keryell: ?

tests/for_each_n.cpp
```diff
@@ -76,21 +76,23 @@ TEST_F(ForEachNAlgorithm, TestStd3ForEachN) {
...
+      val -= adder;
     } else {
-      val = val + adder;
+      val += adder;
     }
+    std::cout << val << std::endl;
   });
-#if PRINT_OUTPUT
+//#if PRINT_OUTPUT
+#if 1
```
keryell: Is this an expected change?

JoanThibault:
There is probably a clean way of merging mapreduce and map2reduce into a mapKreduce using variadic templates and without loosing performances but I did not find it.

keryell:
Thank you for your contribution. It looks good on my side. I give the last word to Codeplay to look if it works with ComputeCpp.
And yes, as you said, there are still some possible code factorization. To be done in the future... :-)

tests/inner_product.cpp
```diff
@@ -146,20 +146,21 @@ TEST_F(InnerProductAlgorithm, TestSycl6InnerProduct) {
 }
 
 TEST_F(InnerProductAlgorithm, TestSycl7InnerProduct) {
-  std::vector<int> v1;
-  std::vector<int> v2;
   int n_elems = 4096;
-  for (int i = 0; i < n_elems; i++) {
+  std::vector<int> v1 (n_elems, 1);
+  std::vector<int> v2 (n_elems, 2);
+  /*for (int i = 0; i < n_elems; i++) {
```
Naghasan: the loop should be deleted, not commented (way better to use the vector ctor)

include/sycl/algorithm/inclusive_scan.hpp
```diff
@@ -107,7 +109,256 @@ OutputIterator inclusive_scan(ExecutionPolicy &sep, InputIterator b,
...
+  q.submit([&] (cl::sycl::handler &cgh) {
+    cl::sycl::nd_range<1> rng
+      { cl::sycl::range<1>{nb_work_group * nb_work_item},
+        cl::sycl::range<1>{nb_work_item} };
+    auto input =
+      input_buffer.template get_access<cl::sycl::access::mode::read>(cgh);
+    auto output =
+      output_buffer.template get_access<cl::sycl::access::mode::write>(cgh);
+
+    cgh.parallel_for_work_group<class wg>(rng, [=](cl::sycl::group<1> grp) {
```
keryell: Ah yes. These are triSYCL-ism.
For example we do not use the template type name of `parallel_for_work_group` for now.

include/sycl/algorithm/find.hpp
```diff
@@ -185,15 +187,17 @@ InputIt find_impl(ExecutionPolicy &snp, InputIt b, InputIt e,
   auto map = [=](size_t pos, value_type x) {
     return (p(x)) ? pos : size;
   };
-  auto red = [=](size_t x, size_t y){return std::min(x, y);};
+
+  auto red = [=](size_t x, size_t y){
```
keryell: Lacking space between `){`.
By the way, is it required to use `red` instead of just passing `std::min` in the next line?

include/sycl/algorithm/count_if.hpp
```diff
@@ -41,7 +41,7 @@
 namespace sycl {
 namespace impl {
 
-#if 0
+#ifdef __COMPUTECPP__
```
keryell: This might not be `__COMPUTECPP__`, but at least Codeplay can change it if they want to try/use the other version.

include/sycl/algorithm/inclusive_scan.hpp
```diff
@@ -181,19 +181,22 @@ void buffer_mapscan(ExecutionPolicy &snp,
...
 
-    cgh.parallel_for_work_group<class wg>(rng, [=](cl::sycl::group<1> grp) {
+    cl::sycl::accessor<B, 1, cl::sycl::access::mode::read_write,
+                       cl::sycl::access::target::local>
+      scratch { cl::sycl::range<1> { size_per_work_group }, cgh };
+
+    cgh.parallel_for_work_group<class wg>(rng_wg, rng_wi, [=](cl::sycl::group<1> grp) {
       //assert(false);
-      B scratch[size_per_work_group];
+      //B scratch[size_per_work_group];
```
keryell: To be removed then.

include/sycl/algorithm/inclusive_scan.hpp
```diff
@@ -181,19 +181,22 @@ void buffer_mapscan(ExecutionPolicy &snp,
...
       input_buffer.template get_access<cl::sycl::access::mode::read>(cgh);
     auto output =
       output_buffer.template get_access<cl::sycl::access::mode::write>(cgh);
 
-    cgh.parallel_for_work_group<class wg>(rng, [=](cl::sycl::group<1> grp) {
+    cl::sycl::accessor<B, 1, cl::sycl::access::mode::read_write,
+                       cl::sycl::access::target::local>
+      scratch { cl::sycl::range<1> { size_per_work_group }, cgh };
+
+    cgh.parallel_for_work_group<class wg>(rng_wg, rng_wi, [=](cl::sycl::group<1> grp) {
```
keryell: Remain the issue of `class wg` which is not unique.
Any idea from Codeplay on how to do this? I guess there is already a way in the PSTL to generate unique class names...

include/sycl/algorithm/inclusive_scan.hpp
```diff
@@ -205,8 +208,8 @@ void buffer_mapscan(ExecutionPolicy &snp,
       // Step 0:
       // each work_item copy a piece of data
       // map is applied during the process
-      grp.parallel_for_work_item([&](cl::sycl::nd_item<1> id) {
-        size_t local_id = id.get_local(0);
+      parallel_for_work_item(grp, [&](cl::sycl::item<1> id) {
+        size_t local_id  = id.get(0) % nb_work_item;
```
keryell: We really need to fix the specification, to avoid this atrocities...

include/sycl/algorithm/inclusive_scan.hpp
```diff
@@ -107,7 +109,258 @@ OutputIterator inclusive_scan(ExecutionPolicy &sep, InputIterator b,
...
+                     size_t size_per_work_group_,
+                     size_t size_per_work_item_,
+                     size_t nb_work_group_,
+                     size_t nb_work_item_):
+    size(size_),
+    size_per_work_group(size_per_work_group_),
+    size_per_work_item(size_per_work_item_),
+    nb_work_group(nb_work_group_),
+    nb_work_item(nb_work_item_) {}
+};
```
Ruyk: I like the idea of using descriptors to have all the preparation of sizes and workgroups in a single place, but seems to me having a different descriptor for all functions is an overkill. Is it not possible to have a single descriptor to dominate them all? Seems most of them have a similar interface, so you could maybe have a sycl_algorithm_descriptor structure defined in the ExecutionPolicy with different constructors for the different use cases?

Ruyk: Also, can you add some documentation explaining what the descriptors are and how should they be used?

include/sycl/algorithm/inner_product.hpp
```diff
@@ -34,7 +34,9 @@
 #include <iostream>
 
 #include <sycl/algorithm/algorithm_composite_patterns.hpp>
+#include <sycl/algorithm/reduce.hpp>
 #include <sycl/helpers/sycl_differences.hpp>
+#include <boost/iterator/zip_iterator.hpp>
```
Ruyk: This causes Parallel STL to require boost when building, which was not there before. Can you make this include only for triSYCL (since I guess triSYCL uses boost so there is no difference)?

include/sycl/algorithm/inner_product.hpp
```diff
@@ -34,7 +34,9 @@
 #include <iostream>
 
 #include <sycl/algorithm/algorithm_composite_patterns.hpp>
+#include <sycl/algorithm/reduce.hpp>
```
Ruyk: The algorithms should not refer to each other, can this be avoided?

include/sycl/algorithm/reduce.hpp
```diff
@@ -94,6 +100,343 @@ typename std::iterator_traits<Iterator>::value_type reduce(
                                      cl::sycl::access::target::host_buffer>();
   return bop(hI[0], init);
 }
+#else
+size_t up_rounded_division(size_t x, size_t y){
+  return (x+(y-1)) / y;
+}
```
Ruyk: This function is defined on the reduce.hpp header, but used on the inclusive_scan.hpp header. This creates a dependency among the two which I am not convinced it is really necessary. If you have common stuff across two algorithms, you can put them on the algorithm_composite_patterns, or you can create your own header with commonalities across triSYCL variants of the implementations.

include/sycl/algorithm/reduce.hpp
```diff
@@ -94,6 +100,343 @@ typename std::iterator_traits<Iterator>::value_type reduce(
...
+ * work properly
+ */
+
+mapreduce_descriptor compute_mapreduce_descriptor(cl::sycl::device device,
+                                                  size_t size,
+                                                  size_t sizeofB) {
+  using std::max;
+  using std::min;
+  if (size <= 0)
+    return mapreduce_descriptor(0, 0, 0, 0);
```
Ruyk: Use { }

include/sycl/algorithm/reduce.hpp
```diff
@@ -94,6 +100,343 @@ typename std::iterator_traits<Iterator>::value_type reduce(
...
+  }
+/*  {
+    auto read_input = input_buff.template get_access
+      <cl::sycl::access::mode::read, cl::sycl::access::target::host_buffer>();
+    for(size_t pos = 0; pos < size; pos++)
+    {
+      std::cout << "[" << pos << "] " << read_input[pos] << ", ";
+    }
+    std::cout << std::endl;
+  }*/
```
Ruyk: can you eliminate this commented code?

include/sycl/algorithm/reduce.hpp
```diff
@@ -94,6 +100,343 @@ typename std::iterator_traits<Iterator>::value_type reduce(
...
+
+  if ((nb_work_item == 0) || (nb_work_group == 0)) {
+    auto read_input = input_buff.template get_access
+      <cl::sycl::access::mode::read, cl::sycl::access::target::host_buffer>();
+    B acc = init;
+    for (size_t pos = 0; pos < size; pos++)
+      acc = reduce(acc, map(pos, read_input[pos]));
+
+    return acc;
+    //return boost::compute::reduce(b, e, init, bop);
```
Ruyk: Can you remove this comment?

include/sycl/algorithm/reduce.hpp
```diff
@@ -94,6 +100,343 @@ typename std::iterator_traits<Iterator>::value_type reduce(
...
+                   cl::sycl::buffer<A, 1> input_buff,
+                   B init, //map is not applied on init
+                   mapreduce_descriptor d,
+                   Map map,
+                   Reduce reduce) {
+
+  size_t size = d.size;
+  size_t size_per_work_group = d.size_per_work_group;
+  size_t nb_work_group = d.nb_work_group;
+  size_t nb_work_item = d.nb_work_item;
```
Ruyk: You create a descriptor but then you extract each variable individually. Can you not just use the descriptor directly?

include/sycl/algorithm/reduce.hpp
```diff
@@ -94,6 +100,343 @@ typename std::iterator_traits<Iterator>::value_type reduce(
...
+
+/*
+ * MapReduce Algorithm applied on a buffer
+ */
+
+template <typename ExecutionPolicy,
+          typename A,
+          typename B,
+          typename Reduce,
+          typename Map>
```
Ruyk: can you explain the different template arguments?

include/sycl/algorithm/reduce.hpp
```diff
@@ -94,6 +100,343 @@ typename std::iterator_traits<Iterator>::value_type reduce(
...
+  }
+/*  {
+    auto read_input = input_buff.template get_access
+      <cl::sycl::access::mode::read, cl::sycl::access::target::host_buffer>();
+    for(size_t pos = 0; pos < size; pos++)
+    {
+      std::cout << "[" << pos << "] " << read_input[pos] << ", ";
+    }
+    std::cout << std::endl;
+  }*/
```
Ruyk: please remove this commented code

include/sycl/algorithm/reduce.hpp
```diff
@@ -94,6 +100,343 @@ typename std::iterator_traits<Iterator>::value_type reduce(
...
+    auto read_input1 = input_buff1.template get_access
+      <cl::sycl::access::mode::read, cl::sycl::access::target::host_buffer>();
+    auto read_input2 = input_buff2.template get_access
+      <cl::sycl::access::mode::read, cl::sycl::access::target::host_buffer>();
+    B acc = init;
+    for (size_t pos = 0; pos < size; pos++)
+      acc = reduce(acc, map(pos, read_input1[pos], read_input2[pos]));
+
+    return acc;
+    //return boost::compute::reduce(b, e, init, bop);
```
Ruyk: please remove this commented code

include/sycl/algorithm/reduce.hpp
```diff
@@ -94,6 +100,343 @@ typename std::iterator_traits<Iterator>::value_type reduce(
...
+ * Map2Reduce on a buffer
+ * */
+
+
+template <typename ExecutionPolicy,
+          typename A1,
+          typename A2,
+          typename B,
+          typename Reduce,
+          typename Map>
```
Ruyk: can you explain the different template parameters?

include/sycl/algorithm/reduce.hpp
```diff
@@ -94,6 +100,343 @@ typename std::iterator_traits<Iterator>::value_type reduce(
...
+      <cl::sycl::access::mode::read>(cgh);
+    auto output = output_buff.template get_access
+      <cl::sycl::access::mode::write>(cgh);
+    cl::sycl::accessor<B, 1, cl::sycl::access::mode::read_write,
+                       cl::sycl::access::target::local>
+      sum { cl::sycl::range<1>(nb_work_item), cgh };
+    cgh.parallel_for_work_group<class wg>(rng, [=](cl::sycl::group<1> grp) {
+      //int sum[nb_work_item];
+      size_t group_id = grp.get(0);
+      assert(group_id < nb_work_group);
```
Ruyk: non-static asserts are not allowed on device kernels

include/sycl/algorithm/reduce.hpp
```diff
@@ -94,6 +100,343 @@ typename std::iterator_traits<Iterator>::value_type reduce(
...
+      B acc = sum[0];
+      for (size_t local_id = 1;
+           local_id < min(nb_work_item, group_end - group_begin);
+           local_id++)
+        acc = reduce(acc, sum[local_id]);
+
+      output[group_id] = acc;
+    });
+  });
+  q.wait_and_throw();
```
Ruyk: Since you have a host accessor inmediatly after, you don't really need the wait and throw. If you want to catch exceptions, you can use throw_asynchronous (which does not wait, only throw exceptions)

include/sycl/algorithm/reduce.hpp
```diff
@@ -94,6 +100,343 @@ typename std::iterator_traits<Iterator>::value_type reduce(
...
+    return init;
+
+  mapreduce_descriptor d =
+    compute_mapreduce_descriptor(device, size, sizeof(value_type));
+
+  auto input_buff = sycl::helpers::make_const_buffer(b, e);
+
+  auto map = [=](size_t pos, value_type x) { return x; };
+
+  return buffer_mapreduce(snp, q, input_buff, init, d, map, bop);
```
Ruyk: is it there a "non-buffer" map reduce?

tests/exclusive_scan.cpp
```diff
@@ -55,7 +55,7 @@ void exclusive_scan_gold(std::vector<T>& v, T init = 0,
 
 // test the gold computation against a known value
 TEST_F(ExclusiveScanAlgorithm, TestSTDExclusiveScan) {
-  std::vector<int> v = {5, 1, 6, 2, 6, 2, 5, 7};
+  std::vector<int> v    = {5, 1, 6,  2,  6,  2,  5,  7};
```
Ruyk: why the extra space?

tests/inclusive_scan.cpp
Ruyk: However, there is no guarantee that the std::plus<int> is valid SYCL device kernel code, and we have found that even trivial things like std::swap can be implemented in very weird ways in some platforms. To avoid introducing accidentally invalid SYCL code from a host library (e.g. system headers), I would rather revert this back to the lambda

tests/exclusive_scan.cpp
JoanThibault: to align the input vector with the with expected output vector

keryell: @Ruyk I thought the same first and I understood the "visual computing" aspect of it with my Emacs and a fixed-space font. :-)

include/sycl/algorithm/inner_product.hpp
keryell: Why should this be an issue, if it is internal?
I am more puzzled by the
```
#include <iostream>
```
why is there a need for I/O in something not related to I/O?

include/sycl/algorithm/inner_product.hpp
keryell: Yes triSYCL uses Boost already.

include/sycl/algorithm/inclusive_scan.hpp
```diff
@@ -107,7 +109,258 @@ OutputIterator inclusive_scan(ExecutionPolicy &sep, InputIterator b,
...
+
+mapscan_descriptor compute_mapscan_descriptor(cl::sycl::device device,
+                                              size_t size,
+                                              size_t sizeofB) {
+  //std::cout << "size=\t" << size << std::endl;
+  using std::min;
+  using std::max;
+  if (size == 0)
+    return mapscan_descriptor(0, 0, 0, 0, 0);
```
keryell: Uniform initialization with { } ?

include/sycl/algorithm/inclusive_scan.hpp
```diff
@@ -107,7 +109,258 @@ OutputIterator inclusive_scan(ExecutionPolicy &sep, InputIterator b,
...
+  size_t local_mem_size =
+    device.get_info<cl::sycl::info::device::local_mem_size>();
+  //std::cout << "local_mem_size=\t" << local_mem_size << std::endl;
+  size_t size_per_work_group = min(size, local_mem_size / sizeofB);
+  //std::cout << "size_per_work_group=\t" << size_per_work_group << std::endl;
+  if (size_per_work_group <= 0)
+    return mapscan_descriptor(size, 0, 0, 0, 0);
```
keryell: Uniform initialization with { } ?

include/sycl/algorithm/inclusive_scan.hpp
```diff
@@ -107,7 +109,258 @@ OutputIterator inclusive_scan(ExecutionPolicy &sep, InputIterator b,
...
+      // Step 0:
+      // each work_item copy a piece of data
+      // map is applied during the process
+      parallel_for_work_item(grp, [&](cl::sycl::item<1> id) {
+        size_t local_id  = id.get(0) % nb_work_item;
+        // gpos: position in the global vector
+        // lpos: position in the local vector
+        for (size_t gpos = group_begin + local_id, lpos = local_id;
+            gpos < group_end;
+            gpos+=nb_work_item, lpos+=nb_work_item) {
```
keryell: Weird spacing around assignment operators. Same is various other places...

include/sycl/helpers/sycl_buffers.hpp
```diff
@@ -138,7 +144,7 @@ reuse_buffer_impl(Iterator b, Iterator e, std::input_iterator_tag) {
  * @param std::input_iterator_tag Used for iterator dispatch only
  */
 template <typename Iterator>
-cl::sycl::buffer<typename std::iterator_traits<Iterator>::value_type, 1,
```
JoanThibault: I removed a trailing space.

include/sycl/algorithm/reduce.hpp
```diff
@@ -94,6 +100,343 @@ typename std::iterator_traits<Iterator>::value_type reduce(
...
+  //std::cout << "local_mem_size=\t" << local_mem_size << std::endl;
+  size_t nb_work_item = min(max_work_item, local_mem_size / sizeofB);
+  //std::cout << "nb_work_item=\t" << nb_work_item << std::endl;
+
+  /* (nb_work_item == 0) iff (sizeof(T) > local_mem_size)
+   * If sizeof(T) > local_mem_size, this means that an object
+   * of type T can't hold in the memory of a single work-group
+   */
+  if (nb_work_item == 0) {
+    return mapreduce_descriptor(size, 0, 0, 0);
```
keryell: guess you can just use everywhere something like:
```
return { size, 0, 0, 0 };
```

include/sycl/algorithm/reduce.hpp
keryell: Yes it would be shorter.

include/sycl/algorithm/reduce.hpp
```diff
@@ -94,6 +100,343 @@ typename std::iterator_traits<Iterator>::value_type reduce(
...
+                   mapreduce_descriptor d,
+                   Map map,
+                   Reduce reduce) {
+
+  size_t size = d.size;
+  size_t size_per_work_group = d.size_per_work_group;
+  size_t nb_work_group = d.nb_work_group;
+  size_t nb_work_item = d.nb_work_item;
+  /*
+   * 'map' is not applied on init
```
keryell: I have the feeling this comment (with a lot of *. Is it the common style in the PSTL?) would benefit to the full function instead.

include/sycl/algorithm/reduce.hpp
```diff
@@ -94,6 +100,343 @@ typename std::iterator_traits<Iterator>::value_type reduce(
...
+                    B init, //map is not applied on init
+                    mapreduce_descriptor d,
+                    Map map,
+                    Reduce reduce) {
+  size_t size = d.size;
+  size_t size_per_work_group = d.size_per_work_group;
+  size_t nb_work_group = d.nb_work_group;
+  size_t nb_work_item = d.nb_work_item;
+  /*
+   * 'map' is not applied on init
```
keryell: Same. Move comment at the top level?

include/sycl/algorithm/reduce.hpp
keryell: Is this can fail?

include/sycl/algorithm/reduce.hpp
```diff
@@ -94,6 +100,343 @@ typename std::iterator_traits<Iterator>::value_type reduce(
...
+
+  if (size <= 0)
+    return init;
+
+  mapreduce_descriptor d =
+    compute_mapreduce_descriptor(device, size, sizeof(value_type));
+
+  auto input_buff = sycl::helpers::make_const_buffer(b, e);
+
+  auto map = [=](size_t pos, value_type x) { return x; };
```
keryell: I have the feeling nothing is captured. So use `[]`

include/sycl/algorithm/reduce.hpp
JoanThibault: no

JoanThibault:
@Ruyk It is correct to assume that everything works with triSYCL. However, I did not manage to make the PSTL works with ComputeCpp. I believe the simplest way is for one person of Codeplay to try it out.

I modified the algorithms so that they can work without having buffer::set_final_data(Iterator) implemented (as this is not part of the specification (yet)).
As it is, the code should work with your implementation, however, if it did not you can define the __USE_OLD_ALGO__ macro which will disable the new algorithms and restore the PSTL in its initial state.
If you define the __TRISYCL__ macro, it means that buffer::set_final_data(Iterator) and creating a buffer from a pair of iterator are implemented.

include/sycl/algorithm/exclusive_scan.hpp
```diff
@@ -148,24 +148,34 @@ OutputIterator exclusive_scan(ExecutionPolicy &snp, InputIterator b,
...
 
     auto d = compute_mapscan_descriptor(device, size - 1, sizeof(value_type));
     buffer_mapscan(snp, q, buffer, buffer, init, d,
                    [](value_type x) { return x; },
                    bop);
   }
 
-  std::advance(o, size - 1);
-  return o;
+  return std::next(o, size -1);
```
keryell: Weird spacing. "size - 1"

CMakeLists.txt
```diff
@@ -24,6 +25,8 @@ else()
   set(CMAKE_CXX_STANDARD 14)
   set(CXX_STANDARD_REQUIRED ON)
 
+  SET (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DTRISYCL_ASYNC=1 -fopenmp -D__TRISYCL__")
```
keryell: Remove `-DTRISYCL_ASYNC=1`.
It does not exist anymore.
What is this `__TRISYCL__` for? When triSYCL is included, it already defines its own flags...

include/sycl/algorithm/exclusive_scan.hpp
```diff
@@ -31,10 +31,13 @@
 
 #include <sycl/helpers/sycl_buffers.hpp>
 #include <sycl/helpers/sycl_namegen.hpp>
+#include <sycl/algorithm/buffer_algorithms.hpp>
 
 namespace sycl {
 namespace impl {
 
+#ifdef __USE_OLD_ALGO__
```
keryell: I would prefer more standard macros with non `__` reserved names, such as `SYCL_PSTL_USE_OLD_ALGO` for example.

include/sycl/algorithm/exclusive_scan.hpp
```diff
@@ -129,6 +132,54 @@ OutputIterator exclusive_scan(ExecutionPolicy &sep, InputIterator b,
...
+          typename OutputIterator,
+          typename T,
+          typename BinaryOperation>
+OutputIterator exclusive_scan(ExecutionPolicy &snp, InputIterator b,
+                              InputIterator e, OutputIterator o, T init,
+                              BinaryOperation bop) {
+
+  cl::sycl::queue q(snp.get_queue());
+  auto device = q.get_device();
+  size_t size = sycl::helpers::distance(b, e);
```
keryell: Is a distance a `std::size_t` and not a `std::ptrdiff_t`? Use `auto` too?

tests/for_each_n.cpp
```diff
@@ -101,7 +101,7 @@ TEST_F(ForEachNAlgorithm, TestSyclForEachN) {
 
 TEST_F(ForEachNAlgorithm, TestSycl2ForEachN) {
   std::array<int, 5> v = {2, 1, 3, 4, 8};
-  std::array<int, 5> result = {3, 2, 4, 5, 9};
+  //std::array<int, 5> result = {3, 2, 4, 5, 9};
```
keryell: Why keeping this?

tests/inclusive_scan.cpp
```diff
@@ -37,9 +37,23 @@ using namespace std::experimental::parallel;
 
 struct InclusiveScanAlgorithm : public testing::Test {};
 
+template<typename T>
+struct plus {
```
keryell: Add some comments about why you define these already existing functors.

tests/transform.cpp
```diff
@@ -45,7 +45,7 @@ TEST_F(TransformAlgorithm, TestStdTransform) {
   std::transform(v.begin(), v.end(), v.begin(),
                  [=](int val) { return val + 1; });
 
-  EXPECT_TRUE(std::equal(v.begin(), v.end(), result.begin()));
+  //EXPECT_TRUE(std::equal(v.begin(), v.end(), result.begin()));
```
keryell: Why this? Where is it tested then?

tests/transform.cpp
```diff
@@ -61,7 +61,7 @@ TEST_F(TransformAlgorithm, TestSyclTransform) {
   sycl::sycl_execution_policy<class TransformAlgorithm2> snp2(q);
   transform(snp2, v.begin(), v.end(), o.begin(),
             [=](int val) { return val + 2; });
-  EXPECT_TRUE(std::equal(o.begin(), o.end(), result.begin()));
+  //EXPECT_TRUE(std::equal(o.begin(), o.end(), result.begin()));
```
keryell: Idem.

tests/transform.cpp
```diff
@@ -72,10 +72,10 @@ TEST_F(TransformAlgorithm, TestSycl2Transform) {
 
   cl::sycl::queue q;
   sycl::sycl_execution_policy<class TransformAlgorithm3> snp(q);
-  int value = 1;
+
   transform(snp, v1.begin(), v1.end(), v2.begin(), o.begin(),
             [=](int val1, int val2) { return val1 + val2 + 1; });
-  EXPECT_TRUE(std::equal(o.begin(), o.end(), result.begin()));
+  //EXPECT_TRUE(std::equal(o.begin(), o.end(), result.begin()));
```
keryell: Idem.

tests/transform.cpp
```diff
@@ -98,5 +98,5 @@ TEST_F(TransformAlgorithm, TestSycl4Transform) {
   sycl::sycl_execution_policy<class TransformAlgorithm4> snp(q);
   transform(snp, v.begin(), v.end(), res_sycl.begin(),
             [=](int val1) { return val1 + 1; });
-  EXPECT_TRUE(std::equal(res_std.begin(), res_std.end(), res_sycl.begin()));
+  //EXPECT_TRUE(std::equal(res_std.begin(), res_std.end(), res_sycl.begin()));
```
keryell: Idem.

Ruyk:
> As it is, the code should work with your implementation, however, if it did not you can define the USE_OLD_ALGO macro which will disable the new algorithms and restore the PSTL in its initial state.
If you define the TRISYCL macro, it means that buffer::set_final_data(Iterator) and creating a buffer from a pair of iterator are implemented.

Did you add this to the Readme?

JoanThibault:
No, but I was waiting for you to try it out, because if it doesn't work I can replace the "USE_OLD_ALGO" by "__COMPUTECPP__" and if does work (depending on the performances), we could remove current algorithms by the new ones.

CMakeLists.txt
```diff
@@ -24,6 +25,8 @@ else()
   set(CMAKE_CXX_STANDARD 14)
   set(CXX_STANDARD_REQUIRED ON)
 
+  SET (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fopenmp -D__TRISYCL__")
```
keryell: Bis repetita: no " -D__TRISYCL__" :-(

benchmarks/nbody.cpp
```diff
@@ -34,6 +34,7 @@
 #include <fstream>
 
 #include <experimental/algorithm>
+#include "CL/sycl.hpp"
```
keryell: Should not this could be automatically included by
```
#include <sycl/execution_policy>
```
?

include/sycl/algorithm/count_if.hpp
```diff
@@ -37,10 +37,13 @@
 #include <sycl/helpers/sycl_buffers.hpp>
 #include <sycl/helpers/sycl_differences.hpp>
 #include <sycl/algorithm/algorithm_composite_patterns.hpp>
+#include <sycl/algorithm/buffer_algorithms.hpp>
 
 namespace sycl {
 namespace impl {
 
+#ifdef SYCL_PSTL_US_OLD_ALGO
```
keryell: To be renamed `SYCL_PSTL_USE_OLD_ALGO` everywhere

include/sycl/algorithm/fill.hpp
```diff
@@ -43,33 +43,31 @@ namespace impl {
...
-      h.parallel_for<typename ExecutionPolicy::kernelName>(
-          r, [aI, val, vectorSize](cl::sycl::nd_item<1> id) {
-            if (id.get_global(0) < vectorSize) {
-              aI[id.get_global(0)] = val;
-            }
-          });
-    };
-    q.submit(f);
-  }
+  cl::sycl::queue q = sep.get_queue();
```
keryell: Bis repetita

include/sycl/helpers/sycl_buffers.hpp
```diff
@@ -74,6 +74,10 @@ template <typename Iterator,
 cl::sycl::buffer<typename std::iterator_traits<Iterator>::value_type, 1>
 make_buffer_impl(Iterator b, Iterator e, std::random_access_iterator_tag) {
   typedef typename std::iterator_traits<Iterator>::value_type type_;
+#ifdef __TRISYCL__
```
keryell: Oh no!

include/sycl/helpers/sycl_buffers.hpp
```diff
@@ -100,14 +105,21 @@ make_buffer_impl(Iterator b, Iterator e, std::random_access_iterator_tag) {
...
 template <typename Iterator,
           typename std::enable_if<
-              !std::is_base_of<SyclIterator, Iterator>::value>::type* = nullptr>
+              !std::is_base_of<SyclIterator, Iterator>::value>::type* =
+                                             nullptr>
 cl::sycl::buffer<typename std::iterator_traits<Iterator>::value_type, 1>
 make_buffer_impl(Iterator b, Iterator e, std::input_iterator_tag) {
-  typedef typename std::iterator_traits<Iterator>::value_type type_;
+  using type_= typename std::iterator_traits<Iterator>::value_type;
+#ifdef __TRISYCL__
```
keryell: Nooooo!

keryell:
Perhaps we can keep the old implementation too.
Can we have for example SYCL_PSTL_USE_OLD_ALGO not set when using triSYCL, so the ComputeCpp can just run as usual in a first phase?

include/sycl/algorithm/fill.hpp
```diff
@@ -45,7 +45,7 @@ namespace impl {
  */
 template <typename ExecutionPolicy, typename ForwardIt, typename T>
 void fill(ExecutionPolicy &sep, ForwardIt b, ForwardIt e, const T &value) {
-  cl::sycl::queue q = sep.get_queue();
+  cl::sycl::queue q { sep.get_queue() };
```
keryell: The usual `auto q =` does not work here?

include/sycl/algorithm/buffer_algorithms.hpp
```diff
@@ -0,0 +1,541 @@
...
+                    sycl_algorithm_descriptor d,
+                    Map map,
+                    Reduce red) {
+    //map is not applied on init
+
+  using std::min;
+  using std::max;
+
+  //WARNING: nb_work_group is not bounded by max_compute_units
+  cl::sycl::buffer<B, 1> scan = { cl::sycl::range<1> { d.nb_work_group } };
```
Ruyk: This constructor is not valid

include/sycl/algorithm/buffer_algorithms.hpp
```diff
@@ -0,0 +1,541 @@
...
+      { cl::sycl::range<1>{ d.nb_work_group * d.nb_work_item },
+        cl::sycl::range<1>{ d.nb_work_item } };
+    auto input = input_buff.template get_access
+      <cl::sycl::access::mode::read>(cgh);
+    auto output = output_buff.template get_access
+      <cl::sycl::access::mode::write>(cgh);
+    cl::sycl::accessor<B, 1, cl::sycl::access::mode::read_write,
+                       cl::sycl::access::target::local>
+      sum { cl::sycl::range<1>(d.nb_work_item), cgh };
+    cgh.parallel_for_work_group<class wg>(rng, [=](cl::sycl::group<1> grp) {
```
Ruyk: The constructor on the specification takes a range, not an nd_range, and there is no implicit conversion among them (This fails to compile in computecpp)

include/sycl/algorithm/find.hpp
```diff
@@ -38,16 +38,20 @@
...
 
+#ifdef __COMPUTECPP__
+
 // a struct to store the result of a predicate comparison, and an index
 // we have to declare this here instead of in the function so that the sycl
 // kernel can see it
 typedef struct search_result {
   bool result;
   int index;
+  search_result() = default;
```
JoanThibault: on line 73, an empty buffer of type search_result is created, without a default constructor for search_result, the creation of this buffer is not possible in a C++ friendly way .

include/sycl/algorithm/transform_reduce.hpp
```diff
@@ -102,6 +105,37 @@ T transform_reduce(ExecutionPolicy& exec, InputIterator first,
...
+
+  auto device = q.get_device();
+  using value_type = typename std::iterator_traits<InputIt>::value_type;
+
+
+  auto d = compute_mapreduce_descriptor(device, size, sizeof(value_type));
+
+  auto input_buff = sycl::helpers::make_const_buffer(b, e);
+
+  auto map = [&](size_t pos, value_type x) { return unary_op(x); };
```
Ruyk: unary_op is here passed by reference, which is not allowed

include/sycl/algorithm/count_if.hpp
```diff
@@ -102,6 +105,36 @@ typename std::iterator_traits<InputIterator>::difference_type count_if(
...
+  auto device = q.get_device();
+  using value_type = typename std::iterator_traits<InputIt>::value_type;
+
+
+  auto d = compute_mapreduce_descriptor(device, size, sizeof(size_t));
+
+  auto input_buff = sycl::helpers::make_const_buffer(b, e);
+
+  auto map = [&](size_t pos, value_type x) {
+    return (unary_op(x)) ? 1 : 0;
```
Ruyk: unary_op is captured by reference, which is not valid

Ruyk:
Seems some kernel has duplicated names, I cannot get it to compile. Can you check the kernel names?

JoanThibault:
Do you have a strategy inside the PSTL to generate a new name for each kernel ?

keryell:
I have the feeling that the PSTL does not provide an automatic way to generate automatically a name and it is up to the user to provide as tempated type to the execution policy, if I correctly understand `include/sycl/execution_policy`.

@Ruyk: do you think it is possible to avoid passing a name every time?

keryell:
I am going on this WIP with branch https://github.com/keryell/SyclParallelSTL/tree/triSYCL because I cannot update this PR.

Ruyk:
Apologies for not merging this branch in. I've been trying to setup a travis testing for both triSYCL and ComputeCpp to make sure all incoming changes are working in both, but I am getting many problems, in both triSYCL and ComputeCpp on the travis machines.
w.r.t the naming, there is a template trick we do to generate names for different types, you can see it in https://github.com/KhronosGroup/SyclParallelSTL/blob/master/include/sycl/helpers/sycl_namegen.hpp Basically, we add a number to each kernel in an algorithm for those algorithms that have multiple kernels. 
However, for each user-defined custom lambda, the user must specify a name via the policy.

keryell:
You do not have to apology because this branch is actually still a WIP.

Yes, I ended up to understand that the kernel naming has to be addressed explicitly at the policy level.

keryell:
I have fixed triSYCL around uninitialized buffers and local accessors, so we should move on this at Xilinx now.
https://github.com/triSYCL/triSYCL/commit/ddc376f89f283ea7864008e43adfcdda8a500781

We will come back with an updated PR.
@a-doumoulakis to check for this.

keryell:
I think this PR should be closed as it has been subsumed by https://github.com/KhronosGroup/SyclParallelSTL/pull/27

PR #28, opened 2017-05-25
Update the Travis config

DuncanMcBain:
Recently the Khronos ICD loader was updated, and rather than fix that
this patch uses the standard OpenCL ICD loader and header packages.
It also updates the config to use ComputeCpp 0.2.0, and adds a fix for
the GCC dual ABI (the 14.04 package is built with GCC 4.9, but this
Travis instance uses 5.4). Additionally, it updates the ComputeCpp
workaround file which had some errors.


DuncanMcBain:
I made some updates to the ComputeCpp SDK repository recently, and Ruyman asked if I could replicate them here.

Ruyk:
Awesome, thanks!

PR #29, opened 2017-05-29
Replace int by actual enum type

Zakor94:
Fixes error with gcc-7.1:
`error: could not convert template argument ‘ModeU’ from ‘int’ to ‘cl::sycl::access::mode {enum}`


Ruyk:
Good catch, thanks

PR #33, opened 2017-06-16
Almost make inner_product compile with sycl buffer

Zakor94:
It doesn't quite compile yet because the execution policy calls inner_product_sequential if the buffer is not a power of 2. The sequential version uses the * operator which is removed for the sycl BufferIterator so it breaks here.
I am not sure what is the cleanest fix for that so I just comment the call to sequential but I don't push that.
I tested the inner_product with small sycl buffers and it works fine.

### DIFF
--- a/include/sycl/algorithm/buffer_algorithms.hpp
+++ b/include/sycl/algorithm/buffer_algorithms.hpp
@@ -287,7 +287,7 @@ B buffer_map2reduce(ExecutionPolicy &snp,
     cl::sycl::accessor<B, 1, cl::sycl::access::mode::read_write,
                        cl::sycl::access::target::local>
       sum { cl::sycl::range<1>(d.nb_work_item), cgh };
-    cgh.parallel_for_work_group<class wg>(rng, [=](cl::sycl::group<1> grp) {
+    cgh.parallel_for_work_group<class wg>(rng.get_global(), rng.get_local(), [=](cl::sycl::group<1> grp) {
       size_t group_id = grp.get(0);
       //assert(group_id < d.nb_work_group);
       size_t group_begin = group_id * d.size_per_work_group;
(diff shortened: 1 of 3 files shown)


include/sycl/algorithm/buffer_algorithms.hpp
```diff
@@ -287,7 +287,7 @@ B buffer_map2reduce(ExecutionPolicy &snp,
     cl::sycl::accessor<B, 1, cl::sycl::access::mode::read_write,
                        cl::sycl::access::target::local>
       sum { cl::sycl::range<1>(d.nb_work_item), cgh };
-    cgh.parallel_for_work_group<class wg>(rng, [=](cl::sycl::group<1> grp) {
+    cgh.parallel_for_work_group<class wg>(rng.get_global(), [=](cl::sycl::group<1> grp) {
```
Zakor94: Maybe there is a better fix here but parallel_for_work_group can't take nd_range. I am not sure if we want to specify a work_group_size as well?

include/sycl/helpers/sycl_iterator.hpp
```diff
@@ -181,6 +181,11 @@ class BufferIterator : public SyclIterator {
     return result;
   }
 
+  BufferIterator<T, Alloc> &operator+=(const int &value) {
```
Zakor94: This is needed for the call to std::next

Ruyk:
I've fixed the conflict with master in your branch. Is this ready to be merged?

Zakor94:
Ha nice. Yes this could be merged but there is still an issue when you try to use inner_product with a sycl::buffer.
It cannot compile because of inner_product_sequential which is called if the size of the buffer is not a power of 2. The error is because inner_product_sequential uses the * operator on a sycl::iterator.

The simple solution I use in my own repository is to comment the call to inner_product_sequential and `assert(false)` instead. This fix makes the tests fail of course. I guess a cleaner solution would be to have a specialisation of inner_product_sequential with sycl::iterator.
So it's up to you to decide what's cleaner.

PR #39, opened 2017-09-12
Implement inner_product_sequential with SYCL

Zakor94:
This is a cleaner pull request than my previous one.

The 2 first commits are leftovers from the previous PR.
The first commit add the operator+= needed when std::next is called, the second one use the local range that has been computed.

The third commit is bigger than expected. As I mentioned in the previous PR inner_product didn't compile when using SYCL buffers because it needed to compile inner_product_sequential when the size is not a power of 2. inner_product_sequential only compiled with std iterators because it used the operator*. I added the implementation for SYCL iterators but the hard part was to select which version to use based on the iterators.

I need all this so that my project can use this repository instead of my own fork!

### DIFF
--- a/include/sycl/algorithm/inner_product.hpp
+++ b/include/sycl/algorithm/inner_product.hpp
@@ -35,29 +35,10 @@
 #include <sycl/algorithm/algorithm_composite_patterns.hpp>
 #include <sycl/algorithm/buffer_algorithms.hpp>
 #include <sycl/helpers/sycl_differences.hpp>
+#include <sycl/helpers/sycl_iterator.hpp>
 
 namespace sycl {
 namespace impl {
-
-// value is true if ClassT has the get_buffer method
-template<class ClassT>
-class has_get_buffer_method {
-private:
-  static auto has_method() {
-    using return_type = decltype(test(&ClassT::get_buffer));
-    return return_type();
-  }
-
-  using MethodReturnT = cl::sycl::buffer<typename std::iterator_traits<ClassT>::value_type, 1,
-                                         typename ClassT::allocator_type>;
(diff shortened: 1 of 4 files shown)


include/sycl/algorithm/inner_product.hpp
```diff
@@ -39,43 +39,85 @@
...
 T inner_product_sequential(ExecutionPolicy &exec, InputIt1 first1,
                            InputIt1 last1, InputIt2 first2, T value,
                            BinaryOperation1 op1, BinaryOperation2 op2) {
-  while (first1 != last1) {
-    value = op1(value, op2(*first1, *first2));
-    ++first1;
-    ++first2;
-  }
-  return value;
+  static constexpr bool UseSycl = has_get_buffer_method<InputIt1>::value && has_get_buffer_method<InputIt2>::value;
```
Ruyk: why do you need this has get buffer? is this to identify the type of the iterator? can it not be detected using the iterator tag?

Zakor94: Hum I didn't notice there was a buffer_iterator_tag I could use. This should be a bit cleaner if I use this but the trick will be the same except that I check if a typedef exists rather than a method.
Or maybe I could just use is_base_of SyclIterator? For some reason I thought I shouldn't include sycl_iterator.hpp but it should be fine right? I'll commit something after lunch.

Ruyk: SyclIterator should be fine to include.

Zakor94:
Alright this is much better with is_base_of. Oh well at least I had fun implementing the other solution!

Ruyk:
Good work, thanks!

PR #41, opened 2017-10-17
CMake generates MSVC 2015 solutions for ParallelSTL 

cjdb:
CMake generates an MSVC 2015 x64 ParallelSTL solution that successfully compiles using the configuration in this branch.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -36,7 +36,10 @@ include_directories("include")
 
 add_subdirectory (src)
 add_subdirectory (examples)
-add_subdirectory (tests)
+
+if (NOT WIN32)
+  add_subdirectory (tests)
+endif()
 
 if (PARALLEL_STL_BENCHMARKS)
   add_subdirectory (benchmarks)
--- a/README.md
+++ b/README.md
@@ -62,14 +62,15 @@ the amount of information copied in and out
 * the ability to specify a queue to the SYCL policy so that the queue is used
 for the various kernels (potentially enabling asynchronous execution of the calls).
 
-Building the project
-----------------------
+## Building the project
 
 This project currently supports the SYCL beta implementation from Codeplay,
-ComputeCPP and the open-source triSYCL implementation.
+ComputeCpp and the open-source triSYCL implementation.
 
 The project uses CMake 3.5 in order to produce build files,
 but more recent versions may work.
+
+### Linux
 
 In Linux, simply create a build directory and run CMake as follows:
 
@@ -81,7 +82,7 @@ In Linux, simply create a build directory and run CMake as follows:
 Usual CMake options are available (e.g. building debug or release).
 Makefile and Ninja generators are supported on Linux.
 
-To simplify configuration, the `FindComputeCpp` cmake module from the ComputeCPP
+To simplify configuration, the `FindComputeCpp` cmake module from the ComputeCpp
 SDK is included verbatim in this package within the `cmake/Modules/` directory.
 
 If Google Mock is found in external/gmock, a set of unit tests is built.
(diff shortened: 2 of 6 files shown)


CMakeLists.txt
```diff
@@ -36,7 +36,10 @@ include_directories("include")
 
 add_subdirectory (src)
 add_subdirectory (examples)
-add_subdirectory (tests)
+
+if (NOT WIN32)
```
Ruyk: Add a comment explaining why tests are disabled on windows

cmake/Modules/FindComputeCpp.cmake
```diff
@@ -232,17 +323,27 @@ endfunction()
...
   # Link with the ComputeCpp runtime library
-  target_link_libraries(${targetName} PUBLIC ${COMPUTECPP_RUNTIME_LIBRARY}
-                        PUBLIC ${OpenCL_LIBRARIES})
-
-endfunction(add_sycl_to_target)
+  target_link_libraries(${targetName} PUBLIC $<$<OR:$<CONFIG:Debug>,$<CONFIG:RelWithDebInfo>>:${COMPUTECPP_RUNTIME_LIBRARY_DEBUG}>
+                                             $<$<NOT:$<OR:$<CONFIG:Debug>,$<CONFIG:RelWithDebInfo>>>:${COMPUTECPP_RUNTIME_LIBRARY}>
+                                             ${OpenCL_LIBRARIES})
 
+endfunction(add_sycl_to_target)
```
Ruyk: Is this different from the module in the computecpp-sdk? If so we can possible update the sdk one?

cmake/Modules/FindComputeCpp.cmake
```diff
@@ -163,29 +202,55 @@ endif()
...
+  set(device_compiler_cxx_standard)
+  get_target_property(targetCxxStandard ${targetName} CXX_STANDARD)
+  if (targetCxxStandard MATCHES 17)
+    set(device_compiler_cxx_standard "-std=c++1z")
+  elseif (targetCxxStandard MATCHES 14)
+    set(device_compiler_cxx_standard "-std=c++14")
+  elseif (targetCxxStandard MATCHES 11)
+    set(device_compiler_cxx_standard "-std=c++11")
+  elseif (targetCxxStandard MATCHES 98)
+    message(FATAL_ERROR "SYCL implementations cannot be compiled using C++98")
```
Ruyk: SYCL implementations -> SYCL applications

Ruyk (changes requested):
Can you update the main Readme file to reflect the support of Visual Studio builds and with instructions on how use it?

cmake/Modules/FindComputeCpp.cmake
cjdb: This is actually copied from the SDK to here!

Ruyk: good then :-)

CMakeLists.txt
cjdb: > Add a comment explaining why tests are disabled ~on windows~

I'm partial to changing this from a Windows thing to all systems. If you don't have the tests pre-installed on _any_ system, then you can't actually run them.

Comment still valid, just needs to be broadened.

README.md
```diff
@@ -135,6 +136,90 @@ features are used, you can ignore those messages as these features do not affect
...
+12. Type 'Bash' and press Enter. Make sure that you aren't opening a mintty client such as Git Bash or Cygwin.
+13. A command prompt window should open, and take you through the final part of the installation process.
+
+#### Install developer tools
+
+1. Type the following commands into Bash. Replace `<username>` with your Windows username.
+
+```bash
+sudo apt update && sudo apt upgrade && sudo apt update
+sudo apt install build-essentials binutils gdb git flex bison texlive-full
```
marios-codeplay: Why do you need flex and bison?

README.md
```diff
@@ -135,6 +136,90 @@ features are used, you can ignore those messages as these features do not affect
...
+mkdir /mnt/c/Users/<username>/projects; cd /mnt/c/Users/<username>/projects
+git clone https://github.com/Kitware/CMake.git
+cd CMake
+./bootstrap && make -j 4 && sudo make install
+cd ..
+sudo apt install ocl-icd-libopencl1 ocl-icd-dev opencl-headers clinfo lsb-core
```
marios-codeplay: clinfo is not required as the user can run computecpp_info to get the OpenCL devices. this is also more relevant as the user can check which devices work with computeCpp

cmake/Modules/FindComputeCpp.cmake
```diff
@@ -97,15 +101,37 @@ else()
...
+  PATHS ${COMPUTECPP_PACKAGE_ROOT_DIR}
   HINTS ${COMPUTECPP_PACKAGE_ROOT_DIR}/lib PATH_SUFFIXES lib
   DOC "ComputeCpp Runtime Library" NO_DEFAULT_PATH)
 
 if (EXISTS ${COMPUTECPP_RUNTIME_LIBRARY})
   mark_as_advanced(COMPUTECPP_RUNTIME_LIBRARY)
-  message(STATUS "libComputeCpp.so - Found")
 else()
-  message(FATAL_ERROR "libComputeCpp.so - Not found!")
+  message(FATAL_ERROR "ComputeCpp Runtime Library - Not found!")
```
marios-codeplay: I would prefer to know what is the file name  as well

marios-codeplay (commented):
Looks good, general comments

DuncanMcBain:
If any comments come up about FindComputeCpp.cmake, could you make a PR to the SDK so we can change it there? Keeps the flow unidirectional which I think will make things a lot cleaner. Thanks!

cjdb:
MSVC++ doesn't support C++11. Pending Clang 6 for C++14, as Microsoft don't support Clang 3.9.

PR #48, opened 2017-12-07
Update Travis Configuration

tremmelg:
* Run in Docker using Ubuntu 16.04
* Updated dependencies for triSYCL (e.g. CMake, boost.compute)
* Use newer compilers: GCC 6 and Clang 4.0


Dockerfile
```diff
@@ -0,0 +1,52 @@
...
+RUN if [ "${c_compiler}" = 'clang-4.0' ]; then apt-get install -yq             \
+    --allow-downgrades --allow-remove-essential --allow-change-held-packages   \
+     clang-4.0 libomp-dev; fi
+
+# GCC 6
+RUN if [ "${c_compiler}" = 'gcc-6' ]; then apt-get install -yq                 \
+    --allow-downgrades --allow-remove-essential --allow-change-held-packages   \
+    g++-6 gcc-6; fi
+
+# OpenCL
```
Ruyk: note this pulls only the ICD loader, not an OpenCL implementation

Dockerfile
```diff
@@ -0,0 +1,52 @@
...
+RUN apt-get install -yq --allow-downgrades --allow-remove-essential           \
+    --allow-change-held-packages ocl-icd-opencl-dev ocl-icd-dev opencl-headers
+
+RUN git clone https://github.com/${git_slug}.git -b ${git_branch} /SyclParallelSTL
+
```
Ruyk: It would be interesting to get pocl and use it to run ComputeCpp on top, but its better if we leave it for a separate issue

cmake/Modules/FindComputeCpp.cmake
```diff
@@ -38,11 +38,6 @@ if(CMAKE_COMPILER_IS_GNUCXX)
     if (CMAKE_CXX_COMPILER_VERSION VERSION_LESS 4.8)
       message(FATAL_ERROR
         "host compiler - Not found! (gcc version must be at least 4.8)")
-    # Require the GCC dual ABI to be disabled for 5.1 or higher
-    elseif (CMAKE_CXX_COMPILER_VERSION VERSION_GREATER 5.1)
-      set(COMPUTECPP_DISABLE_GCC_DUAL_ABI "True")
-      message(STATUS
-        "host compiler - gcc ${CMAKE_CXX_COMPILER_VERSION} (note pre 5.1 gcc ABI enabled)")
```
Ruyk: can you get the latest FindComputeCpp from the computecpp-sdk ? this things are probably fixed already.

.travis/computecpp_workaround.sh
```diff
@@ -5,4 +5,3 @@ mv /tmp/tmp_builtins /tmp/ComputeCpp-latest/include/SYCL/sycl_builtins.h
 
 cat .travis/additional_undef /tmp/ComputeCpp-latest/include/SYCL/host_relational_builtins.h > /tmp/tmp_builtins
 mv /tmp/tmp_builtins /tmp/ComputeCpp-latest/include/SYCL/host_relational_builtins.h
-
```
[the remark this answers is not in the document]
tremmelg: Indeed we don't. I've removed it.

cmake/Modules/FindComputeCpp.cmake
tremmelg: Done.

Dockerfile
tremmelg: Boost.compute (libcompute-dev) also actually pulls them in as it needs the headers.

Dockerfile
tremmelg: It should be possible to automatically install the Intel CPU OpenCL runtime in docker as well.

Ruyk: Not sure if it will work since requires SSE extensions, are they available on travis VM?

tremmelg: I can't find anything on their webpage but there seem to be github projects using SSE instructions.

PR #50, opened 2018-01-02
Add some algorithms

tremmelg:
Added the following algorithms:
- `count`
- `equal`
- `fill_n`
- `mismatch`
- `generate` and `generate_n`
- `reverse` and `reverse_copy`
- `replace` and `replace_if`
- `replace_copy` and `replace_copy_if`
- `all_of`, `any_of` and `none_of`


tests/mismatch.cpp
```diff
@@ -0,0 +1,244 @@
...
+
+TEST_F(MismatchAlgorithm, TestMismatchEqual) {
+  std::vector<int> v{0, 1, 2, 3, 4, 5, 6, 7};
+
+  auto expected = std::mismatch(v.begin(), v.end(), v.begin());
+
+  sycl::sycl_execution_policy<class MismatchAlgorithmEqual> snp{};
+  auto actual = std::experimental::parallel::mismatch(
+      snp, v.begin(), v.end(), v.begin(), v.end(),
+      [](int a, int b) { return a == b; });
```
cjdb: This can be replaced with `std::equal_to<>{}`

tests/mismatch.cpp
```diff
@@ -0,0 +1,244 @@
...
+
+  EXPECT_EQ(actual, expected);
+}
+
+TEST_F(MismatchAlgorithm, TestMismatchMatchCustomPredicate) {
+  std::vector<int> v1{0, 1, 2, 3, 4, 5, 6, 7};
+  std::vector<int> v2{1, 2, 3, 4, 5, 6, 7, 8};
+
+  auto expected = std::mismatch(v1.begin(), v1.end(), v2.begin(),
+      [](int a, int b) { return a < b; });
```
cjdb: `std::less<>{}`

tests/mismatch.cpp
```diff
@@ -0,0 +1,244 @@
...
+
+  EXPECT_EQ(actual, expected);
+}
+
+TEST_F(MismatchAlgorithm, TestMismatchNotEqualCustomPredicate) {
+  std::vector<int> v1{0, 1, 2, 9, 4, 5, 6, 7};
+  std::vector<int> v2{1, 2, 3, 4, 5, 6, 7, 8};
+
+  auto expected = std::mismatch(v1.begin(), v1.end(), v2.begin(),
+      [](int a, int b) { return a < b; });
```
cjdb: `std::less<>{}`

tests/mismatch.cpp
```diff
@@ -0,0 +1,244 @@
...
+
+  sycl::sycl_execution_policy<class MismatchAlgorithmNotEqualCustomPredicate> snp{};
+  auto actual = std::experimental::parallel::mismatch(
+      snp, v1.begin(), v1.end(), v2.begin(), v2.end(),
+      [](int a, int b) { return a < b; });
```
cjdb: `std::less<>{}`

tests/mismatch.cpp
```diff
@@ -0,0 +1,244 @@
...
+
+TEST_F(MismatchAlgorithm, TestMismatchEqualOneRange) {
+  std::vector<int> v{0, 1, 2, 3, 4, 5, 6, 7};
+
+  auto expected = std::mismatch(v.begin(), v.end(), v.begin());
+
+  sycl::sycl_execution_policy<class MismatchAlgorithmEqualOneRange> snp{};
+  auto actual = std::experimental::parallel::mismatch(
+      snp, v.begin(), v.end(), v.begin(), v.end(),
+      [](int a, int b) { return a == b; });
```
cjdb: `std::equal_to<>{}`

tests/mismatch.cpp
```diff
@@ -0,0 +1,244 @@
...
+
+  EXPECT_EQ(actual, expected);
+}
+
+TEST_F(MismatchAlgorithm, TestMismatchMatchOneRangeCustomPredicate) {
+  std::vector<int> v1{0, 1, 2, 3, 4, 5, 6, 7};
+  std::vector<int> v2{1, 2, 3, 4, 5, 6, 7, 8};
+
+  auto expected = std::mismatch(v1.begin(), v1.end(), v2.begin(),
+      [](int a, int b) { return a < b; });
```
cjdb: `std::less<>{}`

tests/mismatch.cpp
```diff
@@ -0,0 +1,244 @@
...
+
+  sycl::sycl_execution_policy<class MismatchAlgorithmMatchOneRangeCustomPredicate> snp{};
+  auto actual = std::experimental::parallel::mismatch(
+      snp, v1.begin(), v1.end(), v2.begin(),
+      [](int a, int b) { return a < b; });
```
cjdb: `std::less<>{}`

tests/mismatch.cpp
```diff
@@ -0,0 +1,244 @@
...
+
+  EXPECT_EQ(actual, expected);
+}
+
+TEST_F(MismatchAlgorithm, TestMismatchNoMatchOneRangeCustomPredicate) {
+  std::vector<int> v1{0, 1, 9, 3, 4, 5, 6, 7};
+  std::vector<int> v2{1, 2, 3, 4, 5, 6, 7, 8};
+
+  auto expected = std::mismatch(v1.begin(), v1.end(), v2.begin(),
+      [](int a, int b) { return a < b; });
```
cjdb: `std::less<>{}`

tests/mismatch.cpp
```diff
@@ -0,0 +1,244 @@
...
+
+  sycl::sycl_execution_policy<class MismatchAlgorithmNoMatchOneRangeCustomPredicate> snp{};
+  auto actual = std::experimental::parallel::mismatch(
+      snp, v1.begin(), v1.end(), v2.begin(),
+      [](int a, int b) { return a < b; });
```
cjdb: `std::less<>{}`

tests/mismatch.cpp
```diff
@@ -0,0 +1,244 @@
...
+
+TEST_F(MismatchAlgorithm, TestMismatchEqualFirstSmaller) {
+  std::vector<int> v{0, 1, 2, 3, 4, 5, 6, 7};
+
+  auto expected = std::mismatch(v.begin(), v.end() - 2, v.begin());
+
+  sycl::sycl_execution_policy<class MismatchAlgorithmEqualFirstSmaller> snp{};
+  auto actual = std::experimental::parallel::mismatch(
+      snp, v.begin(), v.end() - 2, v.begin(), v.end(),
+      [](int a, int b) { return a == b; });
```
cjdb: `std::equal_to<>{}`

tests/mismatch.cpp
```diff
@@ -0,0 +1,244 @@
...
+TEST_F(MismatchAlgorithm, TestMismatchEqualSecondSmaller) {
+  std::vector<int> v{0, 1, 2, 3, 4, 5, 6, 7};
+
+  auto expected = std::mismatch(v.begin(), v.end() - 2, v.begin());
+  std::swap(std::get<0>(expected), std::get<1>(expected));
+
+  sycl::sycl_execution_policy<class MismatchAlgorithmEqualSecondSmaller> snp{};
+  auto actual = std::experimental::parallel::mismatch(
+      snp, v.begin(), v.end(), v.begin(), v.end() - 2,
+      [](int a, int b) { return a == b; });
```
cjdb: `std::equal_to<>{}`

tests/mismatch.cpp
```diff
@@ -0,0 +1,244 @@
...
+
+  EXPECT_EQ(actual, expected);
+}
+
+TEST_F(MismatchAlgorithm, TestMismatchEqualSecondSmaller) {
+  std::vector<int> v{0, 1, 2, 3, 4, 5, 6, 7};
+
+  auto expected = std::mismatch(v.begin(), v.end() - 2, v.begin());
+  std::swap(std::get<0>(expected), std::get<1>(expected));
```
cjdb:
```cpp
using std::swap;
swap(...);
```

(With the Ranges TS, you can just do `ranges::swap(...)`, but that's due to CPOs).

tests/mismatch.cpp
```diff
@@ -0,0 +1,244 @@
...
+TEST_F(MismatchAlgorithm, TestMismatchNotEqual) {
+  std::vector<int> v1{0, 1, 2, 3, 4, 5, 6, 7};
+  std::vector<int> v2{0, 1, 2, 3, 6, 5, 6, 7};
+
+  auto expected = std::mismatch(v1.begin(), v1.end(), v2.begin());
+
+  sycl::sycl_execution_policy<class MismatchAlgorithmNotEqual> snp{};
+  auto actual = std::experimental::parallel::mismatch(
+      snp, v1.begin(), v1.end(), v2.begin(), v2.end(),
+      [](int a, int b) { return a == b; });
```
cjdb: `std::equal_to<>{}`

tests/mismatch.cpp
```diff
@@ -0,0 +1,244 @@
...
+
+TEST_F(MismatchAlgorithm, TestMismatchFirstEmpty) {
+  std::vector<int> v{0, 1, 2, 3, 4, 5, 6, 7};
+
+  auto expected = std::mismatch(v.begin(), v.begin(), v.begin());
+
+  sycl::sycl_execution_policy<class MismatchAlgorithmFirstEmpty> snp{};
+  auto actual = std::experimental::parallel::mismatch(
+      snp, v.begin(), v.begin(), v.begin(), v.end(),
+      [](int a, int b) { return a == b; });
```
cjdb: `std::equal_to<>{}`

tests/mismatch.cpp
```diff
@@ -0,0 +1,244 @@
...
+TEST_F(MismatchAlgorithm, TestMismatchSecondEmpty) {
+  std::vector<int> v{0, 1, 2, 3, 4, 5, 6, 7};
+
+  auto expected = std::mismatch(v.begin(), v.begin(), v.begin());
+  std::swap(std::get<0>(expected), std::get<1>(expected));
+
+  sycl::sycl_execution_policy<class MismatchAlgorithmSecondEmpty> snp{};
+  auto actual = std::experimental::parallel::mismatch(
+      snp, v.begin(), v.end(), v.begin(), v.begin(),
+      [](int a, int b) { return a == b; });
```
cjdb: `std::equal_to<>{}`

tests/mismatch.cpp
```diff
@@ -0,0 +1,244 @@
...
+
+TEST_F(MismatchAlgorithm, TestMismatchBothEmpty) {
+  std::vector<int> v{0, 1, 2, 3, 4, 5, 6, 7};
+
+  auto expected = std::mismatch(v.begin(), v.begin(), v.begin());
+
+  sycl::sycl_execution_policy<class MismatchAlgorithmBothEmpty> snp{};
+  auto actual = std::experimental::parallel::mismatch(
+      snp, v.begin(), v.begin(), v.begin(), v.begin(),
+      [](int a, int b) { return a == b; });
```
cjdb: `std::equal_to<>{}`

tests/none_of.cpp
```diff
@@ -0,0 +1,70 @@
...
+
+  EXPECT_FALSE(result);
+}
+
+TEST_F(NoneOfAlgorithm, TestSyclNoneOfEmpty) {
+  std::vector<int> input{};
+
+  sycl::sycl_execution_policy<class NoneOfAlgorithmEmpty> snp;
+  auto result =
+      none_of(snp, input.begin(), input.end(), [](int) { return false; });
```
cjdb: Please qualify. You can use `namespace parallel = std::experiemntal::parallel;` at the top of this file (and others) if you like.

include/sycl/algorithm/find.hpp
```diff
@@ -93,32 +84,20 @@ InputIt find_impl(ExecutionPolicy &sep, InputIt b, InputIt e,
...
-      return search_result(true, std::min(val1.index, val2.index));
-    } else if (val1.result && !val2.result) {
-      return val1;
-    } else if (!val1.result && val2.result) {
-      return val2;
-    } else {
-      return search_result(false, std::max(val1.index, val2.index));
-    }
-  };
+  auto bop = [=](size_t val1, size_t val2) { return std::min(val1, val2); };
```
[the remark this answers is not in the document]
tremmelg: Complains without the lambda
```
In file included from /home/toomas/projects/SyclParallelSTL/tests/find.cpp:33:
In file included from /home/toomas/projects/SyclParallelSTL/include/sycl/execution_policy:59:
/home/toomas/projects/SyclParallelSTL/include/sycl/algorithm/find.hpp:124:15: error: no matching member
      function for call to 'combine_threads'
            r.combine_threads(std::min<std::size_t>);
            ~~^~~~~~~~~~~~~~~
/home/toomas/projects/SyclParallelSTL/include/sycl/algorithm/algorithm_composite_patterns.hpp:88:8: note:
      candidate template ignored: couldn't infer template argument 'BinaryOperator'
  void combine_threads(BinaryOperator bop) {
       ^
```

include/sycl/algorithm/equal.hpp
```diff
@@ -0,0 +1,151 @@
...
+          cl::sycl::range<1>{std::max(global_size, local_size)},
+          cl::sycl::range<1>{local_size}};
+      auto a1 = buf1.template get_access<cl::sycl::access::mode::read>(h);
+      auto a2 = buf2.template get_access<cl::sycl::access::mode::read>(h);
+      auto aR = bufR.template get_access<cl::sycl::access::mode::read_write>(h);
+      cl::sycl::accessor<int, 1, cl::sycl::access::mode::read_write,
+                         cl::sycl::access::target::local>
+          scratch(cl::sycl::range<1>(local_size), h);
+
+      auto binary_op = [](bool a, bool b) { return a && b; };
```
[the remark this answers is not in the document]
Ruyk: can we guarantee std::logcal_and<>{} does not bring invalid code? how is it usually implemented? I am up for adding it, but my past experiences with other std function implementations (e.g, std::swap) is that sometimes they do clever optimizations for X86 that won't work well on GPUs.

Ruyk:
Please, try to make smaller merge request. I know the code is very similar but is easier to review if it is just one or two functions per pull request.

tremmelg:
@Ruyk I'll keep that in mind. Is it fine if I leave this one or should I break it apart?

include/sycl/algorithm/mismatch.hpp
```diff
@@ -0,0 +1,174 @@
...
+        cl::sycl::helpers::NameGen<0, typename ExecutionPolicy::kernelName> >(
+        r, [a1, a2, aR, length, p](cl::sycl::nd_item<1> id) {
+          size_t m_id = id.get_global(0);
+
+          if (m_id < length) aR[m_id] = p(a1[m_id], a2[m_id]) ? length : m_id;
+        });
+  };
+  q.submit(eqf);
+
+  auto binary_op = [](size_t x, size_t y) { return std::min(x, y); };
```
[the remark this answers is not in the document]
tremmelg: Fails compilation without the lambda.

include/sycl/algorithm/mismatch.hpp
```diff
@@ -0,0 +1,174 @@
...
+  auto input_buff1 = sycl::helpers::make_const_buffer(first1, first1 + length);
+  auto input_buff2 = sycl::helpers::make_const_buffer(first2, first2 + length);
+
+  auto map = [=](size_t pos, value_type1 x, value_type2 y) {
+    return p(x, y) ? length : pos;
+  };
+
+  auto red = [](size_t x, size_t y){
+    return std::min(x, y);
+  };
```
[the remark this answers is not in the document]
tremmelg: Fails compilation without the lambda.

include/sycl/algorithm/equal.hpp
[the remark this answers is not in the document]
cjdb: I think so. See http://eel.is/c++draft/logical.operations#and for its prescribed implementation.

include/sycl/algorithm/find.hpp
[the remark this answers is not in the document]
cjdb: Lambda is fine, we're not doing a reference to a function overload :grin:

include/sycl/algorithm/mismatch.hpp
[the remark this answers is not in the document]
cjdb: Please rename it to `min` then (or if there's only one use of `binary_op`, inline it -- this is preferred).

include/sycl/algorithm/mismatch.hpp
```diff
@@ -0,0 +1,174 @@
...
+template <class ExecutionPolicy, class ForwardIt1, class ForwardIt2,
+          class BinaryPredicate>
+std::pair<ForwardIt1, ForwardIt2> mismatch(ExecutionPolicy& exec,
+                                           ForwardIt1 first1, ForwardIt1 last1,
+                                           ForwardIt2 first2, ForwardIt2 last2,
+                                           BinaryPredicate p) {
+  auto size1 = sycl::helpers::distance(first1, last1);
+  auto size2 = sycl::helpers::distance(first2, last2);
+
+  if (size1 <= 0 || size2 <= 0) std::make_pair(first1, first2);
```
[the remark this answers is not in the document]
cjdb: Ping @tremmelg

include/sycl/algorithm/mismatch.hpp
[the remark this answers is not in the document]
tremmelg: I'll just inline it then.

include/sycl/algorithm/mismatch.hpp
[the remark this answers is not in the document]
cjdb: As before, either inline it (if there's exactly one use-case), or rename it to `min` if there are multiple uses.

include/sycl/algorithm/mismatch.hpp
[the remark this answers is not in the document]
tremmelg: This was supposed to return. Otherwise, it later called `return std::make_pair(std::next(first1, pos), std::next(first2, pos));` with `pos` as 0.

include/sycl/algorithm/equal.hpp
```diff
@@ -0,0 +1,156 @@
...
+            }
+            r.combine_threads(std::logical_and<bool>{});
+            r.workgroup_write_to(aR);
+          });  // end kernel
+    };         // end command group
+
+    q.submit(f);
+    global_size = global_size / local_size;
+    ++passes;
+  } while (global_size > 1);
```
Ruyk: Instead of doing the full reduction, maybe we can "short-circuit" the evaluation.
An additional buffer of size one, initially set to false, can be set to true inside the kernel whenever the predicate is false.
We can copy back to the host this single element at each iteration and check if its true, in which case, all_of is false and we don't need to continue execution.

include/sycl/algorithm/equal.hpp
```diff
@@ -0,0 +1,156 @@
...
+  q.wait_and_throw();
+  auto hr = bufR.template get_access<cl::sycl::access::mode::read,
+                                     cl::sycl::access::target::host_buffer>();
```
Ruyk: use a range accessor to limit the copy back to one element

include/sycl/algorithm/find.hpp
```diff
@@ -29,120 +29,95 @@
...
-          r, [aI, scratch, localRange, length, bop](cl::sycl::nd_item<1> id) {
-            auto r = ReductionStrategy<search_result>(localRange, length, id,
-                                                      scratch);
+          r, [aI, scratch, localRange, length](cl::sycl::nd_item<1> id) {
+            auto r =
+                ReductionStrategy<std::size_t>(localRange, length, id, scratch);
             r.workitem_get_from(aI);
-            r.combine_threads(bop);
+            r.combine_threads([](std::size_t val1, std::size_t val2) {
+              return std::min(val1, val2);
```
Ruyk: use cl::sycl::min

include/sycl/algorithm/find.hpp
```diff
@@ -168,38 +144,32 @@ InputIt find_impl(ExecutionPolicy &sep, InputIt b, InputIt e,
...
-  auto input_buff = sycl::helpers::make_const_buffer(b, e);
+  const auto input_buff = sycl::helpers::make_const_buffer(b, e);
 
-  auto map = [=](size_t pos, value_type x) {
-    return (p(x)) ? pos : size;
-  };
+  const auto pos = buffer_mapreduce(
+      snp, q, input_buff, size, d,
+      [p, size](std::size_t pos, value_type x) { return p(x) ? pos : size; },
+      [](std::size_t x, std::size_t y) { return std::min(x, y); });
```
Ruyk: use cl::sycl::min

include/sycl/algorithm/generate.hpp
```diff
@@ -0,0 +1,78 @@
...
+        cl::sycl::range<1>{std::max(globalRange, localRange)},
+        cl::sycl::range<1>{localRange}};
+    const auto aI = bufI.template get_access<cl::sycl::access::mode::read_write>(h);
+    h.parallel_for<typename ExecutionPolicy::kernelName>(
+        r, [aI, g, vectorSize](cl::sycl::nd_item<1> id) {
+          if (id.get_global(0) < vectorSize) {
+            aI[id.get_global(0)] = g();
+          }
+        });
+  };
```
Ruyk: This will only work on vectors of size up to the maximum number of work items of the device. I think all other algorithms similar to this one are done in the same way. Can we create a separate issue to implement the algorithms so they have multiple passes when there are more items than the number of work items?

include/sycl/algorithm/mismatch.hpp
```diff
@@ -0,0 +1,184 @@
...
+#include <sycl/helpers/sycl_namegen.hpp>
+
+namespace sycl {
+namespace impl {
+
+/* transform_reduce.
+* @brief Returns the transform_reduce of one vector across the range [first1,
+* last1) by applying Functions op1 and op2. Implementation of the command
+* group
+* that submits a transform_reduce kernel.
```
Ruyk: transform_reduce -> mismatch

include/sycl/algorithm/mismatch.hpp
```diff
@@ -0,0 +1,184 @@
+/* Copyright (c) 2015 The Khronos Group Inc.
```
Ruyk: copyright must be updated to 2018 on all files (separate MR please)

include/sycl/algorithm/mismatch.hpp
```diff
@@ -0,0 +1,184 @@
...
+          scratch(cl::sycl::range<1>(local), h);
+
+      h.parallel_for<typename ExecutionPolicy::kernelName>(
+          r, [aR, scratch, passes, local,
+              current_length](cl::sycl::nd_item<1> id) {
+            auto r = ReductionStrategy<std::size_t>(local, current_length, id,
+                                                    scratch);
+            r.workitem_get_from(aR);
+            r.combine_threads(
+                [](std::size_t x, std::size_t y) { return std::min(x, y); });
```
Ruyk: use cl::sycl::min in all kernels

include/sycl/algorithm/mismatch.hpp
```diff
@@ -0,0 +1,184 @@
...
+          });
+    };
+    q.submit(f);
+    ++passes;
+    current_length = current_length / local;
+  } while (current_length > 1);
+  q.wait_and_throw();
+  const auto hR =
+      bufR.template get_access<cl::sycl::access::mode::read,
+                               cl::sycl::access::target::host_buffer>();
```
Ruyk: use a range accessor to copy one element only

include/sycl/algorithm/reverse.hpp
```diff
@@ -0,0 +1,79 @@
...
+        cl::sycl::range<1>{localRange}};
+
+    const auto aI = bufI.template get_access<cl::sycl::access::mode::read_write>(h);
+    h.parallel_for<typename ExecutionPolicy::kernelName>(
+        r, [aI, vectorSize](cl::sycl::nd_item<1> id) {
+          const auto global_id = id.get_global(0);
+          if (global_id < vectorSize / 2) {
+            const auto tmp = std::move(aI[global_id]);
+            aI[global_id] = std::move(aI[vectorSize - global_id - 1]);
+            aI[vectorSize - global_id - 1] = std::move(tmp);
```
Ruyk: can you create a utility function for swap so we can re-use it?  That way is easier to revert back to std::swap in those cases that is known to work

include/sycl/algorithm/equal.hpp
tremmelg: That is undefined behaviour in OpenCL C 2.0 and C++ without synchronisation. I have had failures when threads write the exact same value to the same location without synchronisation on Intel CPUs.

Ruyk: You are correct. Not sure if it will be worse in performance an atomic or the extra kernel(s) launch(s). Ignore my comment for now. Let's keep the algorithms simple: we can always produce optimized versions later.

include/sycl/algorithm/generate.hpp
tremmelg: What do you mean? There is no maximum number of global work-items for a device is there?

Ruyk: There is, and, in some devices, it may be smaller than you think.
From https://www.khronos.org/registry/OpenCL/sdk/1.2/docs/man/xhtml/clGetDeviceInfo.html :

> 
Parameter | Description
-- | --
CL_DEVICE_MAX_WORK_ITEM_SIZES | Return type: size_t[]. Maximum number of work-items that can be specified in each dimension of the work-group to clEnqueueNDRangeKernel.Returns n size_t entries, where n is the value returned by the query for CL_DEVICE_MAX_WORK_ITEM_DIMENSIONS. The minimum value is (1, 1, 1).

tremmelg: But isn't that for a single work-group, so the local size.
*Maximum number of work-items that can be specified in each dimension of the* **work-group**.

Ruyk: You are right, wrong info parameter. Seems it is clearer in CUDA (http://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#compute-capabilities, Maximum x-dimension of a grid of thread blocks, and y and z too).
There is a device param info to query for the maximum global work size *only for built in kernels* (which got me confused):

cl_kernel_work_group_info | Return Type | Info. returned in param_value
-- | -- | --
CL_KERNEL_GLOBAL_WORK_SIZE | size_t[3] | This provides a mechanism for the application to query the maximum global size that can be used to execute a kernel (i.e. global_work_sizeargument to clEnqueueNDRangeKernel) on a custom device given by device or a built-in kernel on an OpenCL device given by device.If device is not a custom device or kernel is not a built-in kernel, clGetKernelArgInfo returns the error CL_INVALID_VALUE.

However, **there is a limit on work item sizes**: the maximum value of size_t *on the device*. From errors in https://www.khronos.org/registry/OpenCL/sdk/1.2/docs/man/xhtml/clEnqueueNDRangeKernel.html:

> CL_INVALID_GLOBAL_WORK_SIZE if global_work_size is NULL, or if any of the values specified in global_work_size[0], ...global_work_size [work_dim - 1] are 0 or exceed the range given by the sizeof(size_t) for the device on which the kernel execution will be enqueued.

Where:

Type in OpenCL C Language | Description | API type for application
-- | -- | --
size_t | The unsigned integer type of the result of the sizeof operator. This is a 32-bit unsigned integer if CL_DEVICE_ADDRESS_BITS defined in clGetDeviceInfo is 32-bits and is a 64-bit unsigned integer if CL_DEVICE_ADDRESS_BITS is 64-bits. | n/a

It is particularly problematic in SYCL when your host is 64 bit and your device is 32, since the maximum array size you can use on the host does not match the device, and you may not get a nice error if that happens!

tremmelg: Oooh, that could be very unexpected. I'll create the issue.

tremmelg: Issue #51

include/sycl/algorithm/reverse.hpp
tremmelg: I get the same compiler error when defining a utility function as when using `std::swap`.
```
In file included from /SyclParallelSTL/tests/reverse.cpp:27:
In file included from /SyclParallelSTL/external/googletest/googlemock/include/gmock/gmock.h:58:
In file included from /SyclParallelSTL/external/googletest/googlemock/include/gmock/gmock-actions.h:43:
In file included from /usr/lib/gcc/x86_64-linux-gnu/5.4.1/../../../../include/c++/5.4.1/algorithm:60:
In file included from /usr/lib/gcc/x86_64-linux-gnu/5.4.1/../../../../include/c++/5.4.1/utility:70:
In file included from /usr/lib/gcc/x86_64-linux-gnu/5.4.1/../../../../include/c++/5.4.1/bits/stl_pair.h:59:
/usr/lib/gcc/x86_64-linux-gnu/5.4.1/../../../../include/c++/5.4.1/bits/move.h:185:11: error: automatic variable qualified with an address space
      _Tp __tmp = _GLIBCXX_MOVE(__a);
          ^
/usr/lib/gcc/x86_64-linux-gnu/5.4.1/../../../../include/c++/5.4.1/bits/algorithmfwd.h:571:5: note: in instantiation of function template specialization 'std::swap<__attribute__((address_space(1))) int>' requested here
    swap(_Tp&, _Tp&)
    ^
1 error generated.
```

```
In file included from /SyclParallelSTL/tests/reverse.cpp:34:
In file included from /SyclParallelSTL/include/sycl/execution_policy:62:
In file included from /SyclParallelSTL/include/sycl/algorithm/reverse.hpp:39:
/SyclParallelSTL/include/sycl/helpers/sycl_swap.hpp:39:5: error: automatic variable qualified with an
      address space
  T tmp = std::move(a);
    ^
/SyclParallelSTL/include/sycl/helpers/sycl_swap.hpp:38:6: note: in instantiation of function template
      specialization 'sycl::helpers::swap<__attribute__((address_space(1))) int>' requested here
void swap( T& a, T& b ) {
     ^
```

tremmelg: Though it works fine if I use `auto tmp = std::move(a);`

keryell:
I have implemented cl::sycl::max and cl::sycl::min for this in triSYCL

keryell:
Otherwise I am not very excited about having our own `sycl::helpers::swap` while some very smart people spent decades devising it... :-/ What is the rationale behind this?

include/sycl/algorithm/reverse.hpp
keryell: Is it a current limitation of ComputeCpp?
It looks like the address space should be dropped when not decorating a reference or a pointer...

PR #53, opened 2018-01-08
[SYCL 1.2.1] Update group::get() to group::get_id()

keryell:
Rename the method to the latest SYCL 1.2.1 standard.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -73,7 +73,7 @@ function install_gmock  {(
 
 function configure  {
   mkdir -p build && pushd build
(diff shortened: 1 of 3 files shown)


keryell:
Ah, ComputeCpp 0.5 is not yet SYCL 1.2.1 on `cl::sycl::group`...

I also need to investigate the issue with Docker, OpenCL & triSYCL...

keryell:
The issue was that after 4 MB of compilation log output, Travis CI just assumes the build is broken :-)

PR #62, opened 2018-02-02
Change methods to inline so as to avoid naming clashes

domantasjurkus:
In some cases, when importing `<sycl/execution_policy>`, a naming clash happens during compilation. Inlining the methods fixes the problem.

### DIFF
--- a/include/sycl/algorithm/buffer_algorithms.hpp
+++ b/include/sycl/algorithm/buffer_algorithms.hpp
@@ -38,7 +38,7 @@ namespace sycl {
 namespace impl {
 
 
-size_t up_rounded_division(size_t x, size_t y){
+inline size_t up_rounded_division(size_t x, size_t y){
   return (x+(y-1)) / y;
 }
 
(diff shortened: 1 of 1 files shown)


include/sycl/algorithm/buffer_algorithms.hpp
```diff
@@ -38,7 +38,7 @@ namespace sycl {
 namespace impl {
 
 
-size_t up_rounded_division(size_t x, size_t y){
+inline size_t up_rounded_division(size_t x, size_t y){
```
cjdb: Let's make this one `constexpr` instead of `inline`.

cjdb:
Looks good, I'll merge after `up_rounded_division` is specified as `constexpr`.

keryell (changes requested):
Yes, please use `constexpr`.
Note that this implies under the hood `inline` that you do not need to repeat.

keryell:
It would be also nice to have a unit test case showing the issue and to check the problem is solved.
Thanks.

keryell:
What is the current mood about this?

keryell:
Any feedback?

PR #68, opened 2018-03-14
Update Travis-CI Docker to Ubuntu 17.10, Clang 5 and GCC 7

keryell:
This is required by the latest triSYCL.
This is a naïve approach since I do not know how to have different Ubuntu versions for ComputeCpp and triSYCL.
So let's try if it breaks ComputeCpp with these versions... :-)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,19 +8,19 @@ services:
   - docker
 
 env:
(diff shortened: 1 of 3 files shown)


keryell:
So it looks like just upgrading the OS and compiler versions is fine for both triSYCL *and* ComputeCpp... :-)

PR #74, opened 2018-05-02
Fix max local size

Rbiessy:
I had an issue with a device that has:
```
Max work item sizes                             256x256x256
Max work group size                             384
```
because SyclParallelSTL assumes that `max_work_group_size <= max_work_item_sizes[i]`

Main noticeable changes are:
* for some algorithm the local size can now be smaller than `max_work_group_size ` if `vectorSize < max_work_group_size`
* public method `sycl_policy::calcultateGlobalSize` no longer exists though I believe it should only be used internally.

PR #75, opened 2018-05-22
Updated to ComputeCpp CE 0.8.0

Rbiessy:


Ruyk:
Seems triSYCL is broken, can you disable triSYCL from travis? Just mark it as optional pass

Rbiessy:
I can't find how to do that :/
On a side note I noticed that CMakeLists.txt adds -DSYCL_PSTL_USE_OLD_ALGO meaning that I actually haven't really tested this changes yet.

Rbiessy:
The tests fail badly when I compile them without SYCL_PSTL_USE_OLD_ALGO, it makes my gpu idle and forces me to reboot my machine. I have the same issue with older commits. Do we want to make this work or should we always use the old algo?

Ruyk:
Keep the USE_OLD_ALGO , the other branch is not fully tested.

keryell:
ah yes, you are using the new 1.2.1 ``h_item`` now... I need to implement them.

Rbiessy:
@keryell There also seemed to have some issues in https://github.com/KhronosGroup/SyclParallelSTL/pull/74 with the `get_info` method: https://travis-ci.org/KhronosGroup/SyclParallelSTL/jobs/382202541

keryell:
Yes, I have to implement this too. Actually it needs some refactoring to be more friendly with virtual inheritance inside triSYCL...

PR #81, opened 2018-07-20
Applied SYCL 1.2.1 renamed member functions for `nd_item` and `nd_range`

cjdb:
* `nd_item::get_local` -> `nd_item::get_local_id`
* `nd_item::get_global` -> `nd_item::get_global_id`
* `nd_range::get_local` -> `nd_range::get_local_range`
* `nd_range::get_global` -> `nd_range::get_global_range`


cjdb:
Failures are for triSYCL only: I assume this means that triSYCL hasn't implemented the changes yet.

keryell:
Actually last week I was thinking why the PSTL + STL was not failing with triSYCL.
In an optimistic mood, I thought this was because there was no impact from SYCL 1.2.1 rev 3 on what is used in the PSTL.
Too optimistic thus. :-)
I will update triSYCL. Thanks for the improvement.

PR #82, opened 2018-08-02
Update benchmarks with new member function names from nd_item to match SYCL 1.2.1 rev 3

keryell:
That is strange, it looks like Travis-CI does not check the compilation of the benchmarks, otherwise this would have been detected...

If you agree with my analysis, please open an issue about this.


keryell:
Ah actually the benchmarks are compiled. So I guess it was working with ComputeCpp because it also support pre-rev 3 SYCL 1.2.1.

The updated triSYCL 1.2.1 is in the reviewing pipeline...

Ruyk:
Yes, we support both on 0.9.1 , issuing a deprecation warning. 
The patch looks fine, thanks!

PR #85, opened 2018-09-24
Fixed: Size of vector checked after buffer construction

Ruyk:
When testing a buffer of invalid size, SYCL buffer throws an
invalid buffer exception.
However, the STL interface should not throw if the size is invalid,
just do a no operation.

This patch reorders the buffer constructor and the vector size check
to ensures invalid SYCL buffers are not created.


AdamHarries:
Looks good to me!

PR #88, opened 2018-12-19
Added implementation & tests for rotate and rotate_copy.

pkeir:


include/sycl/algorithm/rotate_copy.hpp
```diff
@@ -0,0 +1,82 @@
...
+      const size_t rotated_id = (i.get_id(0) + rot_n >= i.get_range(0)) ?
+                                 i.get_id(0) + rot_n  - i.get_range(0)  :
+                                 i.get_id(0) + rot_n;
+      aO[i] = aI[rotated_id];
+    });
+  });
+}
+
+  std::copy_n(data, size, result);
+  delete[] data;
```
keryell: I am a little scarred when I see raw `new`/`delete` in the code and I think to exceptions...
What about just using a `std::vector` created from 2 iterators instead?

pkeir: Fair comment. I've made the change you suggested.

include/sycl/algorithm/rotate_copy.hpp
```diff
@@ -0,0 +1,80 @@
...
+  if (first == last) return result;
+
+  using namespace cl::sycl;
+  using value_type = typename std::iterator_traits<ForwardIt1>::value_type;
+  std::vector<value_type> tmp(first, last);
+  std::copy(first, last, tmp.begin());
+{
+  buffer<value_type> bufI(tmp.cbegin(), tmp.cend()), bufO(tmp.size());
+  bufI.set_final_data(nullptr);
+  bufO.set_final_data(tmp.data());
```
Ruyk: Destination of set_final_data can be an output iterator, so you could avoid the extra copy.
From SYCL 1.2.1:
```cpp
The finalData points to where the outcome
of all the buffer processing is going to be
copied to at destruction time, if the buffer
was involved with a write accessor.
Destination can be either an output iterator
or a weak_ptr_class<T>
```

pkeir: I like that suggestion. The thing is, I get a compilation error (no matching member function) with ComputeCpp 1.0.4 if I provide an iterator to set_final_data. For example, the code below:

```
#include <SYCL/sycl.hpp>
  
int main(int argc, char *argv[])
{
  std::vector<int>      vec(128);
  cl::sycl::buffer<int> buf(128);
  buf.set_final_data(vec.begin());
  return 0;
}
```

Ruyk: That is a good point, we actually missed this variant on the interface. I've reported this internally and I'll keep you posted. In the meantime, I am happy for this to be merged, and possibly and issue be created to re-add the set_final_data with the Output Iterator once it is there.

pkeir: Sounds good. Let me know about set_final_data and I'll add the change then. Cheers.

pkeir: I had a look in ComputeCpp version 1.1.0, but it seems the issue with set_final_data remains. Are you able to merge?

Ruyk: Yes, sure. We are waiting for some more changes into the data model before implementing the missing variant, so we'll re-visit this once its added.

PR #89, opened 2019-04-08
Updated Clang version in Docker and Travis

Ruyk:
* Clang version bumped to 6.0, following triSYCL travis


agozillon:
The set_final_data compile error for triSYCL in job 559.1 of #89 should be fixed when https://github.com/triSYCL/triSYCL/pull/242 is merged into master. 

Not so sure about the Clang-6.0 install failure, I compared the triSYCL docker against the SYCL ParallelSTL's docker. The main differences appear to be that:

- we use ubuntu:bionic rather than ubuntu:artful
- we split the libomp and clang-6.0 install into two separate commands (to support turning OpenMP off/on, rather than avoiding install issues from what I can tell)
- we don't add a repository ppa:ubuntu-toolchain-r/test

I think it's likely that Artful doesn't have clang-6.0 as a package at least from what I can tell: 
https://packages.ubuntu.com/artful/allpackages 

Bionic appears to however: https://packages.ubuntu.com/bionic/allpackages

Artful also appears to have been removed from the packages listing at the end of last year: https://packages.ubuntu.com/ so perhaps a change is in order. 

Hopefully this is somewhat helpful in fixing the problem!

Ruyk:
Thanks!, build is passing now for ComputeCpp and failing due to missing set_final_data on triSYCL.
Are you @agozillon / @keryell  OK for me to merge this even if its failing on triSYCL at the moment?

agozillon:
No problem! I'm fine with merging it, it should be fixed within the next day or two I believe. @keryell ?

keryell:
Thank you for the change.
I am merging since I also merged https://github.com/triSYCL/triSYCL/pull/242

Ruyk:
Thanks! Good team work :-)

